Issue #5, opened 2024-08-13
Expand character data model


stanavdb:
The character data model needs be expanded with several additional fields to hold the following:
- Spheres / currency
- Injuries
- Purpose, obstacle & goals
- Connections
- Bio, appearance & notes

We'll need to think about we model these field

blocks #4

jsc17:
For spheres, the beta rules clearly state the game will only track diamond marks, assuming that your actual wealth is an equally valuable assortment of spheres. Is there any reason to track anything more specific than that for the systems purposes?

stanavdb:
> For spheres, the beta rules clearly state the game will only track diamond marks, assuming that your actual wealth is an equally valuable assortment of spheres. Is there any reason to track anything more specific than that for the systems purposes?

I think we should set it up as generic as possible so that can track various different currencies, all at once if needed. That way we can deal with the various currencies throughout the Cosmere and we're not having to deal with data model migrations later.

Maybe something along the lines of an array field holding the following schema?:
```js
{
    "type": "CURRENCY_TYPE", // The type of currency e.g. Spheres
    "denominations": [
        {
            "id": "DENOMINATION_ID", // The denomination e.g. Marks
            "amount": 100 // The actual amount the character has
        }
    ]
}
```

jsc17:
Yeah, I was overthinking it. I knew it would be simple, but I was wondering if we should include chips, marks, and broams for some reason. The game seems to be fine abstracting it away to just marks, so we should as well. 

The schema looks good unless there ends up being something more nuanced we need to track, but I can't think of anything at the moment.

NoahJGersh:
Injuries are very nebulously defined, which is nice for narrative freedom but tricky for implementation.

Do we just want to track the number of current injuries a character has, in the character schema? We'll probably also want an Injury item type so users can store the duration, location, effect, etc.

NoahJGersh:
Working on some implementation of the currencies specifically, and I'm thinking we might want to shift to a map. It's going to be much easier to establish the data model if we store it as a `SchemaField` rather than an `ArrayField`.

E.g.
```
{
    "spheres": {
        "denominations": [
            {
                "id": "chips",
                "amount": 100
            }
        ]
    },
    "imperials": {
        "denominations": [
            {
                "id": "boxings",
                "amount": 10
            }
        ]
    }
}
```

At that point we could also reassess how we organize the denominations, but this gives us space to derive a `total` for each currency that's computed from the denominations, if we wanted.

[...]

PR #21, opened 2024-08-18
feat(data): add injuries support to character data model

NoahJGersh:
Partially addresses #5 

Adding injury support to the character data model. While injuries themselves (among other conditions) should be Items or some other kind of EmbeddedDocument that lives in a character, we also want easy access to the number of injuries one has among their other conditions. We'll want a derived field for quick access to the number of injuries for Injury Rolls, as well as particular support for Injury Items.

Done:
- Create derived `injuries` field in the common actor data model
- Implement `Injury` item type
  - create additional fields to model an injury (such as `duration`)
- Update Actor to derive `injuries` data from owned Injury items

### DIFF
--- a/src/system/config.ts
+++ b/src/system/config.ts
@@ -106,22 +106,22 @@ const COSMERE: CosmereRPGConfig = {
 
     injuries: {
         [InjuryDuration.FleshWound]: {
-            label: 'Flesh Wound',
+            label: 'COSMERE.Item.Injuries.Duration.FleshWound',
             durationFormula: '1',
         },
         [InjuryDuration.ShallowInjury]: {
-            label: 'Shallow Injury',
+            label: 'COSMERE.Item.Injuries.Duration.ShallowInjury',
             durationFormula: '1d6',
         },
         [InjuryDuration.ViciousInjury]: {
-            label: 'Vicious Injury',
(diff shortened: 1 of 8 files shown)


NoahJGersh:
Work is completed on this PR. There's space for future improvements, like adding active effects support, but that's outside of the scope of this PR and the linked issue

src/system/config.ts
```diff
@@ -103,6 +104,27 @@ const COSMERE: CosmereRPGConfig = {
...
+            label: 'Vicious Injury',
+            durationFormula: '6d6',
+        },
+        [InjuryDuration.PermanentInjury]: {
+            label: 'Permanent Injury',
+        },
+        [InjuryDuration.Death]: {
+            label: 'Death',
+        },
+    },
```
stanavdb, line 126: These labels need to be localized

stanavdb (changes requested):
Looks good! Just one small change for localization.

src/system/config.ts
NoahJGersh: Whoops! Didn't notice that localization is in because I'm forked from main. I'll get those changes in shortly

NoahJGersh:
Alright, we're all localized now after a bit of repo rejiggering to be completely up-to-date with the release branch

stanavdb (approved):

PR #23, opened 2024-08-21
feat(data): add currency support to character model

NoahJGersh:
Partially addresses #5 

Adding currency support to the character model. This is... a tricky one. I settled on a slightly modified config and schema for it, to both be generic and extensible, while still balancing both spheres and boxings in their structures.

This model assumes that most currencies added in the lifetime of the RPG have single denominations, while some (such as spheres) may have an additional axis to that. To support that, all denominations in the `ArrayField` have a mandatory `id`, representing the primary denomination, with an optional `secondaryId` referencing the other denomination. In this way, spheres have gemstones as their primary denomination, with the relative value of each "tier" of gemstone compared to diamonds serving as their conversion value... and they have the actual gem size (chips, marks, broams) as a secondary denomination.

We can then derive the conversionRate of a given denomination field by multiplying the rates of the two constituent denominations. And in the absence of a second denomination, we just go with what's given by the config.

It isn't a perfect system, but it's comprehensive and all based on the config for later expansion.

Done:
- Add currency to Actor model
- Track primary and secondary denominations
- Derive `conversionRate` of a denomination entry from the config, for denominations with one or two ids
- Derive `convertedValue`of a denomination entry from the product of its `conversionRate` and amount
- Derive the total base value of a given currency from its denominations' `convertedValue`s

Incomplete/Out of scope:
- Figure out how to track additional currency-specific qualities
  - e.g. number of infused spheres in a given denomination, which might be useful for Radiants' Investiture Resource
  - Saving this for integration in world models; when individual currencies are being configured for use, we can revisit the schema and find ways to track different values like sphere infused-ness.


NoahJGersh:
While work on this is ongoing, I did make an assumption, that we don't want duplicate tiers. E.g. Sapphire and Amethyst will always be lumped together.

Do we want to just make separate denominations for each kind, and just set them to have equal conversion rates? It might be nice for players who _care_ to specify what kinds of spheres they have in those tiers.

stanavdb:
> While work on this is ongoing, I did make an assumption, that we don't want duplicate tiers. E.g. Sapphire and Amethyst will always be lumped together.
> 
> Do we want to just make separate denominations for each kind, and just set them to have equal conversion rates? It might be nice for players who _care_ to specify what kinds of spheres they have in those tiers.

This is looking great! I'm a little worried about how we get all of this visualised on the character sheet in a sensible manner, but I'm sure we'll figure it out. 

On the point of the duplicate tiers: We should probably just make this a setting so GM's can decide exactly how crunchy they want their game to be. (Perhaps we should even make the whole double denominations for sphere an optional setting and assume diamonds by default).

src/lang/en.json
```diff
@@ -246,6 +246,21 @@
...
+                    "Diamond": "Diamond",
+                    "TierTwoGems": "Garnet/Heliodor/Topaz",
+                    "TierThreeGems": "Ruby/Smokestone/Zircon",
+                    "TierFourGems": "Amethyst/Sapphire",
+                    "Emerald": "Emerald",
+                    "Chips": "Chips",
+                    "Marks": "Marks",
+                    "Broams": "Broams"
+                }
+            }
```
stanavdb, line 263: We need to avoid having any world specifics in the core system. The actual currency specifications will need to be done in the world modules. 

Of course, for local testing its useful to have a currency set up, but let's keep it out of the PR.

NoahJGersh:
> This is looking great! I'm a little worried about how we get all of this visualised on the character sheet in a sensible manner, but I'm sure we'll figure it out.

I considered that. I think having a scrollable div, select menus with the denomination ids, and rows of currency in one's inventory are manageable. Definitely worth some experimentation

> On the point of the duplicate tiers: We should probably just make this a setting so GM's can decide exactly how crunchy they want their game to be. (Perhaps we should even make the whole double denominations for sphere an optional setting and assume diamonds by default).

Sounds like something we would build into each particular world module, then. That way the system can _support_ these things, but the modules themselves will supply the config with the relevant denominations according to the user settings and whatever defaults we want to run with.

I'll go ahead and remove the actual configured denominations and their localization from the PR in the meantime.

NoahJGersh:
Per our discussion on discord, we don't need to store any sort of placeholder type/structure for `Currency`/`CurrencyDenomination`s; we'll leave that sort of structure up to module developers.

src/system/data/actor/common.ts
```diff
@@ -3,6 +3,8 @@ import {
     Resource,
     AttributeGroup,
     Skill,
+    Currency,
+    CurrencyDenomination,
     ExpertiseType,
```
stanavdb, line 8: Looks like there are still references to `Currency` and `CurrencyDenomination` even though they are no longer defined.

src/system/data/actor/common.ts
```diff
@@ -27,6 +29,23 @@ interface ExpertiseData {
     label: string;
 }
 
+interface CurrencyDenominationData {
+    id: CurrencyDenomination;
+    secondaryId?: CurrencyDenomination; // Optional secondary id for doubly-denominated currencies, like spheres
```
stanavdb, line 34: Usage of `CurrencyDenomination`

src/system/data/actor/common.ts
```diff
@@ -52,6 +71,13 @@ export interface CommonActorData {
         { attribute: Attribute; rank: number; mod: Derived<number> }
     >;
     injuries: Derived<number>;
+    currency: Record<
+        Currency,
```
stanavdb, line 75: Usage of `Currency`

stanavdb (changes requested):
Looks good! Just one small change needed.

NoahJGersh:
Fixed the lingering references; we should be good to go now

stanavdb (approved):

PR #25, opened 2024-08-21
chore: fix linting and typing errors

Toby222:
Update dependencies as well (mainly foundry-vtt-types)

D20Roll.toMessage uses the same type signature as Roll.toMessage, which was the only way I could figure out to make TypeScript happy without just telling it to shut up


stanavdb (changes requested):
Could you move the nix changes out into a separate PR? Other than that this looks good

Toby222:
Removed the nix stuff :+1:

stanavdb (approved):

PR #33, opened 2024-08-28
feat: add cosmere initiative system

jsc17:
New PR due to changing to the 0.1.0 branch:

Overrides Foundry's default initiative rolling system to use the Cosmere RPGs slow/fast popcorn style initiative system. Implements a bucket/phases system to group combatants based on turn speed and whether they are a player character or an adversary. Defaults all combatants to slow turn speed when they are added to the combat. Added buttons to the combat tracker allowing them to switch turn speed and indicate they have acted for the turn, and set some basic styling for the combat-tracker template.

Done:

* Create flags on combatant creation to store turn speed and activation status, defaulting to slow for the turn speed.
* Set initiative based on turn speed/character type, instead of rolling
* Add buttons to combat tracker to switch turn speed and indicate the character has activated for the turn (using default foundry icons as placeholders)
* Add headers to the combat tracker to sort player/npc turns (Fast Players, Fast NPCs, etc ...)
* Add context menu option to combat tracker combatants to reset activation status
* Set localization for all text values

### DIFF
--- a/src/system/applications/combat/combat-tracker.ts
+++ b/src/system/applications/combat/combat-tracker.ts
@@ -1,15 +1,19 @@
-import CosmereCombatant from './combatant';
+import { ActorType, TurnSpeed } from '@src/system/types/cosmere';
+import { CosmereCombatant } from '@src/system/documents/combatant';
 
 /**
- * Overrides default tracker template to implement slow/fast buckets and activation button.
+ * Overrides default tracker template to implement slow/fast buckets and combatant activation button.
  */
-export default class CosmereCombatTracker extends CombatTracker {
+export class CosmereCombatTracker extends CombatTracker {
+    // Note: lint rules wants this to be exposed as a readonly field, but base class implements a getter.
     // eslint-disable-next-line @typescript-eslint/class-literal-property-style
     override get template() {
         return 'systems/cosmere-rpg/templates/combat/combat-tracker.hbs';
     }
 
-    //modifies data being sent to the combat tracker template to add turn speed, type and activation status and splitting turns between the initiative phases.
+    /**
+     *  modifies data being sent to the combat tracker template to add turn speed, type and activation status and splitting turns between the initiative phases.
+     */
     override async getData(
         options?: Partial<ApplicationOptions> | undefined,
     ): Promise<object> {
@@ -20,6 +24,7 @@ export default class CosmereCombatTracker extends CombatTracker {
             fastNPC: CosmereTurn[];
             slowNPC: CosmereTurn[];
         };
+        //add combatant type, speed, and activation status to existing turn data.
         data.turns = data.turns.map((turn) => {
             const combatant: CosmereCombatant =
                 this.viewed!.getEmbeddedDocument(
@@ -32,7 +37,7 @@ export default class CosmereCombatTracker extends CombatTracker {
                 turnSpeed: combatant.getFlag(
                     'cosmere-rpg',
                     'turnSpeed',
-                ) as string,
+                ) as TurnSpeed,
                 type: combatant.actor.type,
                 activated: combatant.getFlag(
                     'cosmere-rpg',
@@ -43,35 +48,54 @@ export default class CosmereCombatTracker extends CombatTracker {
             newTurn.css = '';
             return newTurn;
         });
+
+        //split turn data into individual turn "buckets" to separate them in the combat tracker ui
         data.fastPlayers = data.turns.filter((turn) => {
-            return turn.type == 'character' && turn.turnSpeed == 'fast';
+            return (
+                turn.type === ActorType.Character &&
+                turn.turnSpeed === TurnSpeed.Fast
+            );
         });
         data.slowPlayers = data.turns.filter((turn) => {
-            return turn.type == 'character' && turn.turnSpeed == 'slow';
+            return (
+                turn.type === ActorType.Character &&
+                turn.turnSpeed === TurnSpeed.Slow
+            );
         });
         data.fastNPC = data.turns.filter((turn) => {
-            return turn.type == 'adversary' && turn.turnSpeed == 'fast';
+            return (
+                turn.type === ActorType.Adversary &&
+                turn.turnSpeed === TurnSpeed.Fast
+            );
         });
         data.slowNPC = data.turns.filter((turn) => {
-            return turn.type == 'adversary' && turn.turnSpeed == 'slow';
+            return (
+                turn.type === ActorType.Adversary &&
+                turn.turnSpeed === TurnSpeed.Slow
+            );
         });
 
         return data;
     }
 
+    /**
+     * add listeners to toggleTurnSpeed and activation buttons
+     */
     override activateListeners(html: JQuery<HTMLElement>): void {
         super.activateListeners(html);
-        html.find('.cosmere-turn-speed-control').on(
+        html.find(`[data-control='toggleSpeed']`).on(
             'click',
             this._onClickToggleTurnSpeed.bind(this),
         );
-        html.find('.cosmere-activate-control').on(
+        html.find(`[data-control='activateCombatant']`).on(
             'click',
             this._onActivateCombatant.bind(this),
         );
     }
 
-    //toggles combatant turn speed on clicking the "fast/slow" text on the combat tracker window
+    /**
+     * toggles combatant turn speed on clicking the "fast/slow" button on the combat tracker window
+     * */
     protected _onClickToggleTurnSpeed(event: Event) {
         event.preventDefault();
         event.stopPropagation();
@@ -85,7 +109,9 @@ export default class CosmereCombatTracker extends CombatTracker {
         void combatant.toggleTurnSpeed();
     }
 
-    //activates the combatant when clicking the activation icon
+    /**
+     *  activates the combatant when clicking the activation button
+     */
     protected _onActivateCombatant(event: Event) {
         event.preventDefault();
         event.stopPropagation();
@@ -99,7 +125,9 @@ export default class CosmereCombatTracker extends CombatTracker {
         void combatant.setFlag('cosmere-rpg', 'activated', true);
     }
 
-    //toggles combatant turn speed on clicking the "fast/slow" option in the turn tracker context menu
+    /**
+     * toggles combatant turn speed on clicking the "fast/slow" option in the turn tracker context menu
+     */
     protected _onContextToggleTurnSpeed(li: JQuery<HTMLElement>) {
         const combatant: CosmereCombatant = this.viewed!.getEmbeddedDocument(
             'Combatant',
@@ -109,7 +137,9 @@ export default class CosmereCombatTracker extends CombatTracker {
         combatant.toggleTurnSpeed();
     }
 
-    //resets combatants activation status to hasn't activated
+    /**
+     * resets combatants activation status to hasn't activated
+     */
     protected _onContextResetActivation(li: JQuery<HTMLElement>) {
         const combatant: CosmereCombatant = this.viewed!.getEmbeddedDocument(
             'Combatant',
@@ -119,6 +149,9 @@ export default class CosmereCombatTracker extends CombatTracker {
         void combatant.setFlag('cosmere-rpg', 'activated', false);
     }
 
+    /**
+     * Overwrites combatants context menu options, adding toggle turn speed and reset activation options. Removes initiative rolling options from base implementation.
+     */
     _getEntryContextOptions(): ContextMenuEntry[] {
         const menu: ContextMenuEntry[] = [
             {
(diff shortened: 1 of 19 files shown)


src/system/applications/combat/combat-tracker.ts
```diff
@@ -0,0 +1,151 @@
+import CosmereCombatant from './combatant';
+
+export default class CosmereCombatTracker extends CombatTracker {
+    //overrides default tracker template to implement slow/fast buckets and activation button.
+    //eslint-disable-next-line -- eslint rules want readonly field
```
stanavdb, line 5: When using `eslint-disable-next-line` you can specify which rule to disable. While it may not have any further effect here, we should make sure to only disable specific rules when its needed.

src/system/applications/combat/combat-tracker.ts
```diff
@@ -0,0 +1,151 @@
+import CosmereCombatant from './combatant';
+
+export default class CosmereCombatTracker extends CombatTracker {
+    //overrides default tracker template to implement slow/fast buckets and activation button.
```
stanavdb, line 4: This might be nice to include as a doc comment on the whole class

src/system/applications/combat/combat-tracker.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    //modifies data being sent to the combat tracker template to add turn speed, type and activation status and splitting turns between the initiative phases.
+    async getData(
+        options?: Partial<ApplicationOptions> | undefined,
+    ): Promise<object> {
+        const data = (await super.getData(options)) as {
+            turns: CosmereTurn[];
+            [x: string]: unknown;
+        };
+        data.turns = data.turns.map((turn) => {
```
stanavdb, line 18: Is there a reason to type this as `[x: string]: unknown`? 
Couldn't we type the whole thing as:
```ts
turns: CosmereTurn[];
fastPlayers: CosmereTurn[];
slowPlayers: CosmereTurn[];
fastNPC: CosmereTurn[];
slowNPC: CosmereTurn[];
```

Keeping everything properly typed will make it easier for other people to make changes to this code should it be needed in the future.

src/system/applications/combat/combat-tracker.ts
```diff
@@ -0,0 +1,151 @@
...
+            'click',
+            this._onClickToggleTurnSpeed.bind(this),
+        );
+        html.find('.cosmere-activate-control').on(
+            'click',
+            this._onActivateCombatant.bind(this),
+        );
+    }
+
+    //toggles combatant turn speed on clicking the "fast/slow" text on the combat tracker window
```
stanavdb, line 69: This can also be set up as a doc comment

src/system/applications/combat/combatant.ts
```diff
@@ -0,0 +1,42 @@
+import { DocumentModificationOptions } from '@league-of-foundry-developers/foundry-vtt-types/src/foundry/common/abstract/document.mjs';
+import { SchemaField } from '@league-of-foundry-developers/foundry-vtt-types/src/foundry/common/data/fields.mjs';
+import { CosmereActor } from '../../documents';
+
+export default class CosmereCombatant extends Combatant {
+    //sets actor type to CosmereActor, to get typescript to stop yelling. Might be a better way to do this but I'm newish to TS and don't know how.
+    override get actor(): CosmereActor {
+        return super.actor as CosmereActor;
+    }
```
stanavdb, line 9: This is a perfectly fine way to re-declare the type

src/system/applications/combat/combat-tracker.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+}
+
+interface CosmereTurn {
+    id: string;
+    css: string;
+    pending: number;
+    finished: number;
+    type?: string;
+    turnSpeed?: string;
```
stanavdb, line 149: Turn speed should be limited to only be `'fast'` or `'slow'`. You can do this by either making it a literal type or by using an enum.

**Literal**
```ts
type TurnSpeed = 'fast' | 'slow';
```

**Enum**
```ts
const enum TurnSpeed {
    Fast = 'fast',
    Slow = 'slow'
}
```

src/system/applications/combat/combatant.ts
```diff
@@ -0,0 +1,42 @@
...
+        super._onCreate(data, options, userID);
+        void this.setFlag('cosmere-rpg', 'turnSpeed', 'slow');
+        void this.setFlag('cosmere-rpg', 'activated', false);
+        void this.combat?.setInitiative(
+            this.id!,
+            this.generateInitiative(this.actor.type, 'slow'),
+        );
+    }
+
+    generateInitiative(type: string, speed: string): number {
```
stanavdb, line 26: See above about turn speed typing. Actor type enum is available from `@system/types/cosmere`.

stanavdb (changes requested):
No real comments on the implementation. Some minor style/typescript comments & tips. However when I run it locally I get an error when adding a combatant to a combat. 

`Error: An error occurred while rendering CosmereCombatTracker 24. The partial combatant could not be found`

It looks like the combatant partial is not getting preloaded. (Maybe something when wrong when merging with release-0.1.0?) Preloading of partials should be set up in the `preloadHandlebarsTemplates` function, found at: `util/handlebars/index.ts`

src/system/applications/combat/combat-tracker.ts
jsc17: Moved the comment to the whole class, please let me know if you want something more detailed. I'm used to working on solo hobby projects, so commenting is definitely an area I need to work on

src/system/applications/combat/combat-tracker.ts
jsc17: I had it unknown for something I was testing when trying to set it up, fixed

src/system/applications/combat/combat-tracker.ts
jsc17: Not exactly sure what you mean here. Can you elaborate a bit?

src/system/applications/combat/combatant.ts
jsc17: fixed as suggested

jsc17:
Yeah, when switching to the 0.1.0 branch I moved the template into preloadHandleBarTemplates, but didn't commit the updated function. Fixed that and implemented most of the other changes you mentioned.

src/system/applications/combat/combat-tracker.ts
stanavdb: If you use JSDoc comments (comments starting with `/**`) instead of regular comments, then your comment will show up alongside intellisense on most IDEs. Have a look at the Actor document for some examples.

Broadly: Any comments that explain the functionality (for someone who might want to use that function/variable/parameter/etc.) should be doc comments. Any comments that explain the implementation should be regular comments.

src/system/applications/combat/combat-tracker.ts
```diff
@@ -0,0 +1,156 @@
...
+        });
+        data.slowPlayers = data.turns.filter((turn) => {
+            return turn.type == 'character' && turn.turnSpeed == 'slow';
+        });
+        data.fastNPC = data.turns.filter((turn) => {
+            return turn.type == 'adversary' && turn.turnSpeed == 'fast';
+        });
+        data.slowNPC = data.turns.filter((turn) => {
+            return turn.type == 'adversary' && turn.turnSpeed == 'slow';
+        });
```
stanavdb, line 57: These should use the `ActorType` and `TurnSpeed` enums for the comparison. Also its generally best practice to use `===` instead of `==`.

src/system/applications/combat/combat-tracker.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+
+interface CosmereTurn {
+    id: string;
+    css: string;
+    pending: number;
+    finished: number;
+    type?: string;
+    turnSpeed?: string;
```
stanavdb, line 154: Turn speed and type are still typed as string here

src/system/applications/combat/combat.ts
stanavdb, line 1: I don't know how I didn't realise this the first time around, but both `CosmereCombat` and `CosmereCombatant` are Documents in Foundry and should be under the system/documents, not system/applications

src/system/applications/combat/combatant.ts
stanavdb, line 1: See comment combat

src/system/applications/combat/combatant.ts
```diff
@@ -0,0 +1,54 @@
...
+            this.id!,
+            this.generateInitiative(
+                this.actor.type as ActorType,
+                TurnSpeed.Slow,
+            ),
+        );
+    }
+
+    generateInitiative(type: ActorType, speed: TurnSpeed): number {
+        let initiative: number = this.actor.system.attributes.spd.value;
```
stanavdb, line 35: No need to type initiative as `number` here. `actor.system.attributes.spd.value` is already typed as a number so it should infer the type automatically

src/system/applications/combat/combatant.ts
```diff
@@ -0,0 +1,54 @@
...
+        userID: string,
+    ) {
+        super._onCreate(data, options, userID);
+        void this.setFlag('cosmere-rpg', 'turnSpeed', 'slow');
+        void this.setFlag('cosmere-rpg', 'activated', false);
+        void this.combat?.setInitiative(
+            this.id!,
+            this.generateInitiative(
+                this.actor.type as ActorType,
+                TurnSpeed.Slow,
```
stanavdb, line 29: Looking at this I realise I hadn't re-declared the type for actors to ActorType yet. If you add the following to the top of the CosmereActor class you can get rid of all the `as ActorType` casts. 

```ts
    // Redeclare `actor.type` to specifically be of `ActorType`.
    // This way we avoid casting everytime we want to check/use its type
    declare type: ActorType;
```

See system/documents/item for reference.

src/templates/combat/combatant.hbs
```diff
@@ -0,0 +1,50 @@
...
+            <a class="combatant-control {{#if this.hidden}}active{{/if}}" aria-label="{{localize 'COMBAT.ToggleVis'}}"
+                role="button" data-tooltip="COMBAT.ToggleVis" data-control="toggleHidden">
+                <i class="fas fa-eye-slash"></i>
+            </a>
+            <a class="combatant-control {{#if this.defeated}}active{{/if}}"
+                aria-label="{{localize 'COMBAT.ToggleDead'}}" role="button" data-tooltip="COMBAT.ToggleDead"
+                data-control="toggleDefeated">
+                <i class="fas fa-skull"></i>
+            </a>
+            {{/if}}
```
stanavdb, line 16: This doesn't appear to work for me. I'm connected as a GM user but I don't see the visibility toggle. Maybe you need to `@root.user.isGM`? Or perhaps the context isn't getting pass through properly because it's a partial.

src/templates/combat/combatant.hbs
```diff
@@ -0,0 +1,50 @@
...
+    </div>
+    {{/if}}
+    <div class="cosmere-combatant-buttons">
+        <a class="cosmere-turn-speed-control"></a>
+            {{#if this.activated}}
+            <a class="cosmere-activate-control-activated"></a>
+            {{else}}
+            <a class="cosmere-activate-control"></a>
+            {{/if}}
+    </div>
```
stanavdb, line 49: Having tried it, it would be nice to have some tooltips (using `data-tooltip` attr) that indicate what the buttons do

stanavdb (changes requested):
Noticed a few more things that I'd missed/hadn't been able to check yet last time.

src/system/applications/combat/combat-tracker.ts
```diff
@@ -0,0 +1,156 @@
+import CosmereCombatant from './combatant';
+
+/**
+ * Overrides default tracker template to implement slow/fast buckets and activation button.
+ */
+export default class CosmereCombatTracker extends CombatTracker {
+    // eslint-disable-next-line @typescript-eslint/class-literal-property-style
```
stanavdb, line 7: As a good habit, when I disable an eslint rule I add a note comment (`// NOTE: ...`) describing why that rule had to be disabled. This is useful for when other people look at the code (or yourself after a long while away) and forces you to really justify needing to disable eslint rules. 
Disabling eslint rules should be generally discouraged, but is sometimes needed

src/system/applications/combat/combat-tracker.ts
jsc17: Fixed enums and replaced all ==

src/system/applications/combat/combat.ts
jsc17: I was following along with a different modules structure and didn't really know the difference. From what I understand now, documents are data (like actors) and applications will have ui components (like the character sheets). I know there's more to it than that, but is that essentially correct?

src/templates/combat/combatant.hbs
jsc17: This appears to have been something between them being in an each block and a partial. Adding @root to the appropriate sections resolved the issue.

src/system/applications/combat/combat-tracker.ts
jsc17: This was a pretty good explanation. I hadn't used JSDoc comments before so I didn't realize they did that for intellisense. I've expanded the comments in the code I've added.

stanavdb (approved):

Issue #47, opened 2024-09-21
Style adversary sheet


stanavdb:
Same as #44 but for the adversary sheets.

# Design
![n01_SheetMockups_Actions](https://github.com/user-attachments/assets/44c80a26-4069-4d0b-b4e2-e56dbfc331fe)
![n02_SheetMockups_Equipment](https://github.com/user-attachments/assets/51bdd87d-3867-426c-b7f2-f837d8acc0de)
![n03_SheetMockups_Effects](https://github.com/user-attachments/assets/3cc9f7c2-2db3-405d-91df-d55bf4214703)

zithith:
see https://github.com/stanavdb/cosmere-rpg/issues/44#issuecomment-2408226253

PR #78, opened 2024-10-18
Characters

### LINKED ISSUE #44: Style character sheet
The current design for the character sheet is functional, but doesn't yet match the awesome designs by @MangoFVTT. 
The general structure is mostly there, but a lot of clean up on the CSS/HTML is needed.

# Designs
01_SheetMockups_Details
02_SheetMockups_Actions
03_SheetMockups_Equipment
04_SheetMockups_Goals
05_SheetMockups_Effects

### LINKED ISSUE #71: Add theme support
We want to support a different themes for sheets (Actor/Item, possibly other UI elements as well) for all the various Cosmere worlds that will be released. These themes will be provided through their respective world modules. 

A robust theme system that allows us to provide themes in a standardised way through world modules. Users will need to be able to select their preferred themes, likely with some defaulting if only a single world module is used. 

**Checklist:**

PilotKayer:
**Type**  

- [ ] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  
_A clear and concise description of the changes you are making. What does this PR implement or fix?_

CSS and HTML updates on both Adversary and Character sheets to better align with what was proposed by MangoVFTT. It is going for a more clean look with less visual clutter. See screenshot attached.

**Related Issue**  
_This PR must be linked to an existing issue. Please specify the related issue number and whether this PR fully or partially resolves it. Use the format `Closes #<issue-number>` or `Partially addresses #<issue-number>`._

Partially closes #44 and #47 . Closes #71 .

**How Has This Been Tested?**  
_Describe how you tested the changes and how others can replicate the testing steps._

image

**Checklist:**  
- [X] I have tested my changes on Foundry VTT version: `Version 12 Stable - Build 331`.

**Additional context**  
Colours setting have been updated to work with variables. This means that it should fully support themes and therefore should be able to close #71.
I have made changes in the "default" colours and look from what was originally proposed to avoid visual clutter. This sheets are game pieces that need to be clean and clear to understand. I agree that making them pretty was necessary to "bling" the system. However, the proposed sheets had too much details and too much use of the accent colour. I scaled down the use of special SVGs and the accent colour, and instead opted for a clean look to make it a more clear game piece rather than something which simply looks good.


src/style.scss
```diff
@@ -25,6 +25,18 @@
...
+    --plotweaver-color-accent: #ffcb74;
+    --plotweaver-color-faded: #f6f6f6;
+    --plotweaver-color-sheet: #111111;
+
+    --plotweaver-color-base-1: #000000;
+    --plotweaver-color-base-2: #1f1f1f;
+    --plotweaver-color-base-3: #2f2f2f;
+    --plotweaver-color-base-4: #3f3f3f;
+    --plotweaver-color-base-5: #4f4f4f;
+    --plotweaver-color-base-6: #5f5f5f;
```
MangoFVTT, line 38: There's already a `&.theme-plotweaver` as the default theme, why add a whole new set?

src/style.scss
```diff
@@ -16,7 +16,7 @@
     --plotweaver-color-grey-4: #444;
     --plotweaver-color-grey-5: #555;
     --plotweaver-color-white: white;
-    --plotweaver-color-off-white: #f0f0e0;
+    --plotweaver-color-off-white: #fffff0;
```
MangoFVTT, line 19: This should not be changed as it is the colour used by Foundry itself for showing settings icons and such and we don't want mismatches

TheFlyingPig:
This is some great work!  I think MangoFVTT is right, we should stick as close as possible to their proposed design, and then for changes we should discuss and debate them in Discord or on an issue here.

src/style.scss
PilotKayer: To put it very simply, it doesn't work. I have attached an image to show how it looks if I comment out this section.
image

src/style.scss
PilotKayer: Good catch, it has been reverted.

src/style.scss
zithith: I suspect that's because we may have some dodgy scss syntax? though I feel like it should work.
Either way we'd need to make sure the theme class is applied to the correct element, which I'm not sure we do at the moment? It may be hard coded to something right now for dev though, so have a look. Might want to look at how various other modules handle theme switching? especially tying it to a world setting (I'll try looking up some of the Star Wars UI ones in this regard tomorrow and link back).
For now, this link might help? https://bcodes.io/blog/post/theming-css-variables-in-scss

MangoFVTT: If the default theme doesn't work then the stormlight theme is unlikely to either, so I would figure out getting it to work before we commit anything since themes are very important to how the premium modules will be set up

PilotKayer: @zithith it is not to do with the CSS structure, but to how the root is not receiving the theme tag. Furthermore, @MangoFVTT, I might have to update this PR to not include issue #71 as I unfortunately do not have the time to debug the theming issue.
My current priority is making sure that the SVGs, colours and positions are correct.

zithith: @PilotKayer Yeah there's definitely something to look into here, and I think it relates to the fact we are relying on default foundry dark mode to handle a lot for us in the base sheets.
In release 0.1.0 I added in a setting toggle for item sheet tabs that required inserting a class higher up the chain than the sheets themselves could handle, so we probably need to do something similar for the theme switch. Happy to pick up #71 myself and try this approach (after we've figured out what glaring parts of the 0.1 release need immediate attention)

PilotKayer:
### Draft 2 release for the Character sheet.
This draft has been built to better represent the original design by @MangoFVTT .
#### To Note:
1. Adversary sheet is currently broken in this build
2. I have left in the gradients - please let me know if I should leave it in or remove it until we figure out how to add the suggested "HD pack"
3. This has not been implemented for "light mode", and is, therefore, broken.

Screenshots:
image
image
image
image
image

src/style/sheets/actor/character.scss
```diff
@@ -164,18 +274,95 @@
...
             flex-direction: column;
 
-            border: 1px solid #463a47;
-            overflow: hidden;
-            background: #0b0a13;
+            background: var(--plotweaver-color-base-2);
             text-shadow: 0 0 0.5rem black;
             box-shadow: 0 0 0.5rem black;
 
+            .border {
```
zithith, line 286: if this border block is exactly the same as the one above (or very similar to it and/or other declarations for borders) could we maybe make a SCSS mixin?

zithith (approved):
Overall comments:
This is great and I think we want to get it in soon, but some comments for potential future passes:
* There's a lot of hard-coded paths in the html templates, ones that I suspect are the same?... can we find a way to inject these dynamically at least, so that there is less maintenance overhead. and to potentially also optimise the final markup too.
* These CSS files are getting HUGE. I know it wasn't part of this issue, but we probably want a ticket to spend time really thinking about how we can reduce duplication of definitions and optimise this, both for the SCSS in the repo and the size of the outputted native CSS

MangoFVTT (approved):
I agree with Zitith, I'm worried about all the hardcoded paths in the templates because it means we can't easily change SVGs if we want to. It's gonna be hell to maintain if we ever need to change layouts and sizes of borders. Other than that though this is fine. I'll do a clean-up pass on the css when I implement colours for light themes

PR #108, opened 2024-10-23
Allow Damaging Items to Override Graze Damage

### LINKED ISSUE #102: Add an option to damage rolls to freely customise graze damage
Currently all damage rolls (as part of an attack) must use skill/attributes mod for graze. This works fine for character but adversaries are often non-standard and don't fit into this paradigm.

Add some options to use custom graze damage in the form of a formula, instead of the default calculated. 

-

**Version**  
0.1.1

**Additional context**  
Reported by `@doc_sun` on Discord.

**Checklist:**

zithith:
**Type**  

- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
Allow damaging effects to specify a new formula for calculating graze damage value. This is mostly to satisfy the requirement for NPCs/Adversaries who can have values in official statblocks for grazing hits that don't follow the standard rules for attacks.

**Related Issue**  
Closes #102

**How Has This Been Tested?**  
_Describe how you tested the changes and how others can replicate the testing steps._
In Progress

**Checklist:**  
- [x] I have tested my changes on Foundry VTT version: [insert version here].

**Additional context**  
I will presume that this change won't add any sorts of flags or complexity such as a checkbox to add ability mod to the damage or the like. If GMs want an odd action that uses a mod in the grazing formula, then they can just use the `@` property shorthand?


src/style/sheets/item/module.scss
```diff
@@ -304,6 +304,29 @@ app-item-properties {
...
+        .header {
+            display: flex;
+            align-items: center;
+
+            label {
+                font-weight: bold;
+                flex: 2;
+            }
+
+            span {
```
zithith, line 318: [ ] to remove span styling

src/templates/item/components/details-damage.hbs
```diff
@@ -61,4 +61,37 @@
...
+        <input name="system.damage.grazeOverrideFormula"
+            type="text"
+            value="{{item.system.damage.grazeOverrideFormula}}"
+            {{#if (not editable)}}
+            readonly
+            {{/if}}
+        >
+    </div>
+    {{/if}}
+</div>
```
stanavdb, line 96: Graze is only relevant in the case of a skill test, right? I think this should be moved into the  `hasSkillTest` if-statement above.

stanavdb:
Can you drop a comment in #102 so that I can assign it to you?

src/templates/item/components/details-damage.hbs
zithith: Yeah, now I think about it you're right. That's a little implicit in the way the rules describe it so didn't stand out to me late last night 😂
But I can't see an example where the official stats or special rules suggest this can be done outside of a roll. I think there are effects that remove the focus cost to do it? But not anything that allows you to apply graze damage without making an attack roll

zithith:
update: https://discord.com/channels/1299110557689053264/1300257113766891533/1300257113766891533

src/system/documents/item.ts
```diff
@@ -569,7 +597,7 @@ export class CosmereItem<
      */
     public async use(
         options: CosmereItem.UseOptions = {},
-    ): Promise<D20Roll | [D20Roll, DamageRoll] | null> {
+    ): Promise<D20Roll | [D20Roll, ...DamageRoll[]] | null> {
```
MangoFVTT, line 600: It may be more flexible to our future needs for use to just always return an array of `[ Roll ]` (with no specific types of rolls needed to be specified). We may want to have utility rolls that return a formula that isn't a d20 or a damage roll for example. I think it's also better to just have it be a single array of roll objects instead of nested arrays too, because it's easier to filter to get what we want that way.

src/system/documents/item.ts
```diff
@@ -555,12 +583,12 @@ export class CosmereItem<
                 user: game.user!.id,
                 speaker,
                 content: `<p>${flavor}</p>`,
-                rolls: [skillRoll, damageRoll],
+                rolls: [skillRoll, damageRolls],
```
MangoFVTT, line 586: Similar to my other comment I think it's better here to just collate all rolls into a flat array of rolls because it'll be way easier to manipulate when creating the chat message styling. Some chat message properties are also already set up with a flat array in mind (e.g. the damage rolls property on a chat message searches the array for all objects instanceof DamageRoll, and is not expecting DamageRol[])

src/system/documents/item.ts
```diff
@@ -375,12 +375,40 @@ export class CosmereItem<
         // Perform the roll
         const roll = await damageRoll(
             foundry.utils.mergeObject(options, {
-                formula: this.system.damage.formula,
+                formula: `${this.system.damage.formula}+${rollData.mod}`,
```
stanavdb, line 378: If we're going to include the mod in the formula, shouldn't we just require that its actually included in the formula, instead of adding it like this? That would give more freedom for homebrewers, but it would make extracting the base damage harder.

src/system/documents/item.ts
```diff
@@ -375,12 +375,40 @@ export class CosmereItem<
...
                 damageType: this.system.damage.type,
                 mod: rollData.mod,
                 data: rollData,
+                source: this.name,
             }),
         );
+        // We want to store the unmodded damage for use in graze calcs
+        // This isn't a particularly perfect solution, but it's functional
+        // only undoing the automatic addition of the selected attribute
+        rollData.baseRoll = (roll.total ?? 0) - (rollData.mod ?? 0);
```
stanavdb, line 388: Maybe we should pass through a couple of different options, like:
- `@damage.value` - The total damage that was rolled, including mod
- `@damage.unmodded` - The total damage without the modifier
- `@damage.dice` - The total of all the dice (excludes any additions)

I'm not attached to the names, though I do think `baseRoll` isn't super clear.

src/system/documents/item.ts
```diff
@@ -375,12 +375,40 @@ export class CosmereItem<
...
+        const grazeFormula =
+            this.system.damage.grazeOverrideFormula ?? `${rollData.baseRoll}`;
+        if (grazeFormula) {
+            grazeRoll = await damageRoll(
+                foundry.utils.mergeObject(options, {
+                    formula: grazeFormula,
+                    damageType: this.system.damage.type,
+                    data: rollData,
+                }),
+            );
```
stanavdb, line 401: Since we're passing the base roll results as the formula for the graze, will that end up showing up in the chat message? I wonder if it's perhaps better to pass the actual evaluated terms? Or the damage formula (without mod) but then apply the dice results to the dice terms?

src/system/documents/item.ts
zithith: So I did debate making it just a list of Rolls. that would indeed be easier to manage in most areas of the code, and more flexible in future.
Mostly I went this way because I was following Stan's lead, because I also realised that if I swapped to a Roll array, I'd also need to include some degree of identification process to find where various rolls are (to be fair, this should probably also be done in this current set up, but the compile-time checking makes me feel like it is less necessary and we can assume the index locations are right).
I don't know just how many times we'll want the automated roller in the system to be able to handle anything that falls outside of the "Skill roll + potential damage roll" baseline that makes up the majority of Plotweaver's interaction mechanic. It would be good to be ready for it, but is it just belt-and-braces right now?
I'd like a way that "just works" regardless of what is thrown at it, but we'd need more supporting code around that I feel?

As for the nested structure, at this point (and also where it is passed to the chat message) it is becoming un-nested. The "tuple with spread operation" is the only way I could find to define an array where certain types are at certain indexes while still being undefined length. I could have gone with `(D20Roll | DamageRoll)[]` but at that point it might as well be Roll[], this is basically saying the format will be `[D20Roll, DamageRoll, DamageRoll, DamageRoll, ...]`

src/system/documents/item.ts
zithith: Ah, I meant to add the spread operator in here.
Will do a quick push

zithith: done, silly thing was the damage-only branch was already doing this 🤦🏻‍♂️

src/system/documents/item.ts
zithith: I see where you're coming from, but the flipside of such flexibility is that all the standard items then have to actively declare it. Potentially we could achieve the same level of flexibility by providing a checkbox to add the attribute value and default it to on?
Seems like it's a fair assumption for the system to apply this core mechanic to damaging effects (unless this extra modifier only applies to attacks, and other damaging items don't apply this? In which case, default the checkbox to off).
Certainly if I was playing an early release of a system, I'd not be upset that it was making such an assumption even if it meant I couldn't model a particular item I was homebrewing.

src/system/documents/item.ts
zithith: Good thinking.
I guess the simple way would be to set the graze roll's dice property to that of the standard damage roll? That seems like it would be a solution about as simple as current? But I don't know if that actually works?
However I'm wondering if this is potentially just going to introduce unnecessary levels of branching, i.e. we'd need to first check if there isn't an override formula or the formula includes "the base roll string" then the formula used would have to get set based on which of those it was and the previous roll's dice inserted. Otherwise we just use the override formula as is?
I think the current approach covers the existing requirements and in conjunction with @MangoFVTT 's design discussions on discord, regardless of which damage pot you have selected for applying, when you expand the accordion to see the dice details it'll be showing both the default roll and the graze formula anyway.

src/system/documents/item.ts
zithith: I thought the baseRoll term might be a bit confusing, though I think whatever terminology we use it will require some level of explanation (hence I included it in the hint text). I do quite like the idea of storing it in an object with multiple options though...

src/system/documents/item.ts
MangoFVTT: Why not pass the full roll object as `baseRoll`, then take the terms of baseRoll, remove the mods, and do Roll.fromTerms() on the remainder to get another roll object that we can pass as `grazeRoll`?  For non-override I mean.

zithith: Yeah, that sounds positive, I don't really know the Roll API well enough to remember what's possible.
So if we have someone who wants to write an override formula for like the cremlings, say, which is `floor(@baseRoll/2)` will the foundry engine pick up the total from the roll object correctly, or will it need to evaluate to a numeric term?

MangoFVTT: If we don't have a dice breakdown we can just not show the tooltip, which is fine. Obviously most people would prefer to HAVE the tooltip always, but if you use the Roll API that shouldn't be an issue. It does all the logic of evaluating terms internally afaik, including Function terms like the one you gave

src/system/documents/item.ts
stanavdb: We should either auto include the mod (as its currently set up to do), or not include it (so it needs to be explicitly added to the formula), anything in between is less clear imo. I think its fine if all standard items have to declare it, I believe that's what 5e does as well. And anyone homebrewing is going to look at the standard items to figure out how to set up their own stuff.
If we auto include the mod there's just certain things that we won't be able to model. I imagine there might be adversaries that can't be easily set up without it.

Also from just a general clarity point; If something is labelled "formula" it should be the formula. The formula shouldn't then be it plus something else. At least not without a hint telling you.

MangoFVTT: I agree with requiring the mod to me manually entered. It's how it's set up in other d20 systems as well

zithith: Disclaimer: I apologise if my tone in any of this sounds short or harsh, I've just dealt with the mountain/molehill dilemma of getting my nearly 3 year old to bed and I am low on spoons for double checking my sentences. But I do want to find some time to update this with whichever approach we favour tomorrow, so want the discussion to continue. I mean anything said here to be constructive and honestly just see if I have missed something in my understanding.

If you guys prefer to set it to manual, that's fine. Thinking about it, whichever way we implement it, there will be a set of users who assume it's the other way or don't know and just roll something and see how it works - or look at pre-existing examples as you say - in which case it's an easy fix, especially if they have access to a compendium of such examples. I *guess* the expectation for damage values tend to be less for the base roll automation as it is for the basic skill roll; though it's nice if there's an easier way to have it magically do it with few clicks in my experience. And I still am unsure how well-versed the average VTT user is with roll formulas, especially where there's anything platform specific, without a helper/guide. It was doing macros and modding MapTool over 15 years ago that really made me consider programming as a career, and for about half the time since I've been involved with VTTs but I still find my players (who also GM other groups on VTTs) struggle to write out a fairly basic dice pool 🤷🏻‍♂️
If we do go this route however I question whether we need to bother setting a damage stat at all then? might as well just add all the actor attributes to the context the formula needs and then the user can put `@spe` or `@str` or whatever. Alternatively we keep automatic mod addition, but as I suggested above, make it toggleable from the item details window (i.e. a checkbox for "Ignore Modifier?" and have it default to on unless you're creating a weapon?).

> either auto include the mod (as its currently set up to do), or not include it (so it needs to be explicitly added to the formula), anything in between is less clear imo.

Just for my clarity, I believe that the approach I have here *is* an implementation of auto-including the mod. It's not in the middle, I just chose to apply the mod on the roll here rather than in the apply damage step. I feel that made more logical sense as to how I think of the math when doing the rolls manually at an in-person table; it matches closer to how the skill/attack roll is calculated and also pulls similar logic into a closer location in the code, you don't have to go dig to find out where the mod is being added.
If I've missed some implication that makes this different, or even wrong by the rules, I'm sorry I missed it, please let me know 👍🏻

zithith: (Side Note: I just realise I forgot to remove the apply mod from the damage application in the chat cards as I was telling myself to leave those bits for Mango's changes, so we're doubling the mod on application at the moment! I'll be sure to resolve that one way or another once this is decided!😅)

MangoFVTT: The point IS to allow users who want to to add @spe or @str or whatever they want to the damage formula. By default, we should support @mod to add whatever attribute is used for the skill test, but we shouldn't be hardcoding it to ONLY allow that. A checkbox is fine, but it means if someone wants to skill test with STR then add INT to the damage (for example), they have to first remove the checkbox then add @int manually to the formula. Essentially, like @stanavdb said, the "formula" field becomes not actually the formula, but _part_ of the formula. To me that's even more confusing to setup than just writing "1d6 + @mod" in the damage. 

I do agree that the mod should be included in the actual damage roll object, and not just added later. As I've mentioned elsewhere, my ideal solution for damage rolls would be to have a nested graze damageroll object that can be called directly, so that when a user pressed apply damage we just go "ok full damage is selected, use the base damage roll" or "ok graze damage is selected, use the graze damage roll" and not have to be doing a bunch of logic checks in apply damage functions.

zithith: unless I'm mistaken by the item setup, we already do separate the damage attribute (red) from the attack one (yellow), and just set the default for the damage to use what is selected for the activation?
image

I don't know why we have a skill for the damage entry, but I suppose someone might need to use it for something later on.
My comment earlier was that if we don't want to add it automatically we should drop the select boxes in the damage area and just add instructions that explains how to access the attributes/skill ranks through @-notation in the formula; or we keep as is and just add a button to turn off the automatic addition (with `@mod` also being available as a shorthand for the attack modifier so that if you wanted to change both together it would be one input change later).
----
I was trying to get this changeset to end up with the nested roll as you suggested, how far off from what you are after am I? (assuming that the other discussion threads about how we generate the roll values doesn't change the structure of what's done so far, just the values)

MangoFVTT: The object structure atm seems fine to me and in line with how I expected it to be set up. I was just re-iterating that that should stay the case. 

TBH i forgot that we have attribute drop downs in the damage details. I think the sensible thing would be to remove that and have @mod specifically refer to the attribute from the skill test. Then if someone wants to add custom attributes, they can just @ that specific attribute as normal.

stanavdb: The attribute dropdown is relevant with regards to graze damage. Surely we only want the mod to drop out for graze damage and not any other flat modifiers? This way the user can target what they want the `@mod` value to be, which is what gets dropped for graze, but still apply any other flat damage based on the roll data.

zithith: *noises of consternation*
you're right. Assuming we're decided on implementing graze as just "hit damage total - main attribute value". But if we're also deciding that the user has to manually call out adding the damage mod, do we then insist they call out a graze override when they use a mod other than the attack mod?
Or going even further, just ask them to define the graze formula all the time and not have a default for them to override?

stanavdb (approved):

PR #117, opened 2024-10-26
Centralise and Localise Settings

MangoFVTT:
**Type**  
- [ ] Bug fix
- [ ] Feature
- [X] Refactor

**Description**  
Refactor the Settings script to do a couple of things:
- Create a single "getSettingValue" function that we can call without needing to import the SYSTEM ID everywhere. Also, if the foundry API for getting a setting value ever changes (which it has done in the past), we can simply change it in one place and be done.
- Create an object reference of Setting Names that can be imported elsewhere to refer to a specific setting directly without hardcoding the setting name everywhere. Also used to localise the setting key in the lang file, though unfortunately it has to be hardcoded in the lang file, so if a setting name changes we have to manually change it there.
- Setup for each loops for settings batches so the settings script isn't a huge list of duplicate code blocks
- Localise setting names and hints in the settings registry so that they can be used in different languages.

**Checklist:**  

**Additional context**  
As the settings list slowly expands, it will be very useful to manage setting names and setup in one singular place. Especially if we ever need to rename a specific setting. (Ignore the Skip dialog setting for now, that's just so I can have multiple batches set up for testing. I will use that later)

Also, setting localisation is 100% necessary so I set that up at the same time.

### DIFF
--- a/src/system/settings.ts
+++ b/src/system/settings.ts
@@ -4,7 +4,7 @@ import { SYSTEM_ID } from './constants';
  * Enumerable of identifiers for setting names.
  * @enum {String}
  */
-export const SETTING_NAMES = {
+export const SETTINGS = {
     INTERNAL_FIRST_CREATION: 'firstTimeWorldCreation',
     INTERNAL_LATEST_VERSION: 'latestVersion',
     ITEM_SHEET_SIDE_TABS: 'itemSheetSideTabs',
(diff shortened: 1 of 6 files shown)


src/system/settings.ts
```diff
@@ -1,26 +1,73 @@
...
+
+/**
+ * Index of identifiers for system settings.
+ */
+export const SETTINGS = {
+    INTERNAL_FIRST_CREATION: 'firstTimeWorldCreation',
+    INTERNAL_LATEST_VERSION: 'latestVersion',
+    ITEM_SHEET_SIDE_TABS: 'itemSheetSideTabs',
+    ROLL_SKIP_DIALOG_DEFAULT: 'skipRollDialogByDefault',
+};
```
zithith, line 11: suggest making this `as const`

zithith: actually, could you structure it something like:
```
{
    internal: {
        FIRST_CREATION: '...',
        ...
    },
    items: {
        SHEET_SIDE_TABS: '...',
        .... ,
        defaults: {
            side_tabs: false
        }
    },
    rolls: {
        SKIP_SOMETHING: '...',
    }
}
```
Hopefully this could make the looping script even more repeatable?

PR #121, opened 2024-10-27
Add Roll Modifier Keys and Config Skipping

### LINKED ISSUE #114: Make Rolling Without a Dialog the Default
One of the major points of annoyance in the dnd5e system on Foundry is the amount of clicks one needs to go through in order to get to a chat message, particularly for DMs that are rolling a lot of characters all at once. This problem is way less for this system, but every roll having a dialog popup to raise stakes and all that will still, in the long run, be a LOT of extra clicks. The majority of times, the roll that people want will be the default configuration of the dialog, so needing to accept that dialog is a redundant extra click.

I would like to propose the following workflow for a regular roll:
-  If the user simply clicks on a skill or action, immediately roll with the default roll configuration. No raised stakes, no advantage/disadvantage, just a simple basic roll (we can make this behaviour swappable with the "show dialog" behaviour via a setting if desired)
- If the user Shift-clicks (we can and should set this up so the key modifiers are all customisable) a skill or action that has a roll, set the d20 to Advantage and otherwise keep the default configuration of everything. There's always a d20 in a test, so it is by definition the most common target for advantage. Shift-clicking something that has no roll simply doesn't do anything different than a normal click.
- If the user Ctrl-clicks a skill or action, same as above, but disadvantage.
- If the user R-clicks (or some other key, doesn't matter) a skill or action, raise the stakes on the roll and otherwise keep the default configuration of everything
- If the user Alt-clicks a skill or action, show the full roll dialog and allow 100% customisation of the roll as it is now (or vice versa if the setting proposed above is a thing we want)

In this way, we cover the bases for all the most common types of roll, while only requiring a single click for the user to get what they want. The dialog remains an option for comprehensive customisation of the roll.

The simple way to do this would be to pass advantage/disadvantage/raise stakes as config options from a preliminary function that detects if key modifiers are pressed when the click is made.  The config should also have a fastForward value that is set to true by default and false if the relevant modifier is clicked. This also opens up these options for macros in future, so a user can simply call with the relevant config options to get what they want without even click modifier keys.

**Additional context**  
I have some history with this from the dnd5e side, as developer for [Ready Set Roll](https://github.com/MangoFVTT/fvtt-ready-set-roll-5e), which a lot of people consider a must have to skip the billion dialogs that dnd5e forces on you. What I'm proposing is in-line with the behaviour that module creates for rolling workflow, so it is a very tried and tested method.

MangoFVTT:
**Type**  

- [ ] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  
Implements all the functionality for system keybindings that modify a roll and skip its configuration dialog when clicking, as discussed on Discord per issue #114. This includes functionality to swap between "skip dialog by default" and "show dialog by default" as per the system setting added in PR #117.

**Related Issue**  
Closes #114 

**How Has This Been Tested?**  
Tested all combinations of key modifiers together, and tested with the Skip Dialog by Default setting enabled/disabled. All rolls behave correctly with or without key modifiers. This includes item rolls.

image

**Checklist:**  
- [X] I have tested my changes on Foundry VTT version: 12.331

**Additional Context:** 
I should have set up things in such a way that, when we add macro support, config parameters can be hardcoded into the call and propagate correctly down to override key modifiers and the like.


MangoFVTT:
Leaving this as draft for now because it depends on #117 to merge first.

stanavdb (changes requested):
Testing locally, with the dialog on by default. When I use an item with an attack, it gives me the attack dialog, and when I click roll, it gives me the dialog again but without the damage. Click roll on the final dialog does perform the roll.

image
image

MangoFVTT:
@stanavdb Fixed, was a variable rename I accidentally committed that wasn't caught by the typing 👍

src/system/utils/index.ts
stanavdb, line 1: Everything looks good, just one minor thing I want to raise; I think we should be exporting the actor utils as `ActorUtils` instead of directly. (And just generally, anything that isn't in generic should be exported with a name imo.) 
That way we avoid confusion with certain util functions only working for actors even though they're coming from the utils import.

stanavdb (approved):

src/system/utils/index.ts
MangoFVTT: If we're going to do that I vote we make every utils file its own Utils object then. So even generic would be GenericUtils, just to keep things consistent.

stanavdb: Another alternative, we could also just not provide an index for the utils folder. That way you're always forced to specifically import what you need.

MangoFVTT: That's a good solution yeah, I'll change to that

PR #142, opened 2024-10-31
Fixed ContextMenu Errors from issue 104

### LINKED ISSUE #104: Contextmenu error in console
**Describe the bug**  
When opening certain Application (character/adversary sheet for example) an error appears in the console. This does not seem to cause any actual problems, though we might simply be overlooking them.

**To Reproduce**  
1. Load up a world
2. Open a character sheet
3. Close the character sheet
4. Open up the character sheet again
5. Check the console for the error

**Expected behavior**  
No error should be thrown.

**Screenshots**  
image

**Foundry VTT Version**  
12.331

0.1.1

**Checklist:**

tbytnar:
**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
Contextmenu error in console whenever a character sheet is reopened.  

**Related Issue**  
Closes #104 

**How Has This Been Tested?**  
Opening, closing, then reopening any character sheet any number of times.  Since this symptom led from issues with Effects, Actions, and Equipment I also added several of each of these to character sheets and confirmed that they still render after my changes.

N/A

**Checklist:**  

**Additional context**  
Delayed Initialization: Since the context menu setup now waits for rendering, any code that relies on the context menu being ready sooner might need to be adjusted, though I could not find anything that suggested this would be an issue..

Re-rendering Events: This hook will trigger only once per rendering cycle.  I think this is okay, but would rather defer to someone with more expertise here.  Are there any dynamic elements that might depend on this?

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -77,6 +77,10 @@ function markdownParser(config) {
 
                 // Construct the destination path
                 const dest = path.join(srcPath, config.targets[index].dest, `${fileName}.html`);
+                const destDir = path.join(srcPath, config.targets[index].dest);
+                if(!fs.existsSync(destDir)){
+                    fs.mkdirSync(destDir);
+                }
 
                 // Write the parsed markdown to the destination path
                 fs.writeFileSync(dest, `<div>${markdown}</div>`);


stanavdb:
Could you leave a comment on #104? That way I can assign you to the issue.

tbytnar:
> Hey! thanks for the PR. I think there may be some missing commits? There's no change that actually triggers the hook?
> 
> In case there's no missing commits: The component system is not native to foundry, I built it as I didn't like the default handlebars application implementation. It doesn't have hooks, yet (practically everything in the system lacks hooks at this time).
> 
> However I think that regardless this isn't actually a solution. If you have a look at the context menu class, you can see that it binds to the component's render event. Meaning the context menu's render happens after the component's render already.

Ahhh, so all I've done is effectively disabled the context menus.  Sorry I'm very new to this world and have a lot to learn.

stanavdb:
> Ahhh, so all I've done is effectively disabled the context menus. Sorry I'm very new to this world and have a lot to learn.

No worries! Issue #104 isn't the easiest to pick up for anyone not familiar with the code base. Do you want me to unassign you from the issue, or do you want to keep working on it?

tbytnar:
Keep me on it.  This is how I learn.  I'm a glutton for punishment.

I know the error messages stem from the create() method in AppContextMenu because the parent (AppContextMenu.Parent) being passed in doesn't have (an accessible) element.  I have a workaround:

`        const checkElement = () => {
            if (this.element) {
                AppContextMenu.create(
                    this as AppContextMenu.Parent,
                    'right',
                    // (menu items here)
                );
            } else if (attempts < 10) {
                attempts++;
                setTimeout(checkElement, 100); // Retry after 100ms
            } else {
                console.error("Failed to initialize AppContextMenu: element is not available.");
            }
        };
        checkElement();`

But I dont like it, has to be a better way and that's what I'm digging through now.

PR #144, opened 2024-11-02
Issue 104 Fix.  ListComponent timing change and cleanup methods implemented.

tbytnar:
…ted.

**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
The ListComponents were calling the .create() method of AppContextMenu before their own element objects were fully initialized.  This caused the error as .AppendChild was clearly not available inside of the ContextMenu create() method.  I've changed this functionality to take place in _onRender instead of _onInitialize to ensure better timing of the element initializations.  I also implemented some cleanup methods _onDestroy in the ListComponents which properly cleans up their ContextMenus, Thanks Claude!

**Related Issue**  
Fixes Issues #104 

**How Has This Been Tested?**  
Opening, closing and opening a character sheet any number of times.  Also adding objects to each of the lists still saves and reloads on subsequent restarts of Foundry.

**Checklist:**  

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -77,6 +77,10 @@ function markdownParser(config) {
 
                 // Construct the destination path
                 const dest = path.join(srcPath, config.targets[index].dest, `${fileName}.html`);
+                const destDir = path.join(srcPath, config.targets[index].dest);
+                if(!fs.existsSync(destDir)){
+                    fs.mkdirSync(destDir);
+                }
 
                 // Write the parsed markdown to the destination path
                 fs.writeFileSync(dest, `<div>${markdown}</div>`);
--- a/src/system/applications/actor/components/actions-list.ts
+++ b/src/system/applications/actor/components/actions-list.ts
@@ -162,6 +162,7 @@ export class ActorActionsListComponent extends HandlebarsApplicationComponent<
     static TEMPLATE =
         'systems/cosmere-rpg/templates/actors/components/actions-list.hbs';
 
+    private contextMenu?: AppContextMenu;
     /**
      * NOTE: Unbound methods is the standard for defining actions
      * within ApplicationV2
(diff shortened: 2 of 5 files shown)


tbytnar:
@stanavdb this feels more comprehensive of a fix, but thoughts on the change from _onInitialize to _onRender?

stanavdb:
I don't think that should solve the issue, or at least it doesn't address the underlying problem. Any fix will probably have to happen in either the [context menu class](https://github.com/stanavdb/cosmere-rpg/blob/main/src/system/applications/utils/context-menu.ts) or the [component system](https://github.com/stanavdb/cosmere-rpg/blob/main/src/system/applications/component-system/system.ts). 

The create method of the context menu already binds to the render event of the component:
```ts
// Create context menu
const menu = new AppContextMenu(parent, anchor, items);

// Add event listener
parent.addEventListener('render', async () => {
    await menu.render();
    menu.bind(...selectors);
});

return menu;
```

Also there is no `_onDestroy` for components. See the [component base class](https://github.com/stanavdb/cosmere-rpg/blob/main/src/system/applications/component-system/component.ts) for what you can hook into.

stanavdb:
Oh I think I've accidentally stumbled on the bug while find the code to link. The context menu calls its render method in its construct, even though it should wait until its parent has rendered. I think that line can just be entirely removed. (And the constructor should be made private)

tbytnar:
> Oh I think I've accidentally stumbled on the bug while finding the code to link. The context menu calls its render method in its construct, even though it should wait until its parent has rendered. I think that line can just be entirely removed. (And the constructor should be made private)

Okay, THAT makes a lot of sense.  Sorry for the hassle on this one, but I appreciate the guidance and patience with me.  Updated PR incoming.

stanavdb (approved):

stanavdb:
Can you change the base branch to `release-0.1.2`. We're keeping the main branch in line with current release

PR #146, opened 2024-11-04
feat(talent trees): add basic talent trees

### LINKED ISSUE #143: Add support for talent trees
Currently talent trees are implicit; Talents either belong to a path or ancestry and have some prerequisites. However this solution doesn't scale, as new sorts of talent trees are required (for example surge talent trees).

Add a new generic "Talent Tree" item type, for managing talent trees. With this the Specialty item can be removed (/migrated to the talent tree item).

Using a talent tree mixin that can be applied to specialties, surges, etc. This also limits scaling as the talent trees are limited to the existing item types. Homebrewers won't be easily able to add new talent trees that don't neatly fit into the existing categories. 

**Version**  
0.1.1

**Checklist:**

stanavdb:
**Type**  
- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
This PR adds Talent Tree items, which serve as the root for talents/advancement. Specialty items are replaced with the more generic Talent Trees.

**Related Issue**  
Closes #143 

**How Has This Been Tested?**   
Created a new Talent Tree item called spy. Dragged on the talents from the compendium. Then switched the talent tree to view mode. Added the opportunist key talent to my test actor. Set skills to the appropriate values required for the first talents. Checked if the "Sure Outcome" and "Plausible Excuse" talents had the correct borders (indicating they are available), and clicked them to ensure they were added to the actor correctly.

image
image
image
image

**Checklist:**  

---

### To do
- [x] Make individual nodes editable
- [x] Make connections removable
- [x] Add tooltip with talent info on hover
- [x] Add functionality to create new connection between existing nodes
- [x] Make talent tree resizable
- [x] Replace embedded document field with document uuid field
- [x] Make connections derived from prerequisites
- [x] Prevent adding the same item twice
- [x] Register talent tree with each item for updating
- [x] Add ability to pick a talent from the tree


src/system/utils/handlebars/index.ts
```diff
@@ -490,6 +501,7 @@ export async function preloadHandlebarsTemplates() {
         'systems/cosmere-rpg/templates/item/goal/partials/goal-details-tab.hbs',
         'systems/cosmere-rpg/templates/item/power/partials/power-details-tab.hbs',
         'systems/cosmere-rpg/templates/item/path/partials/path-details-tab.hbs',
+        'systems/cosmere-rpg/templates/item/talent-tree/partials/talent-tree-node-tooltip.hbs',
         'systems/cosmere-rpg/templates/combat/combatant.hbs',
         'systems/cosmere-rpg/templates/chat/parts/roll-details.hbs',
         'systems/cosmere-rpg/templates/chat/parts/chat-card-header.hbs',
```
MangoFVTT, line 507: Note that if we merge #151 and this PR around the same time, we need to move this to the new format for template registration, which is putting the path into the TEMPLATES object so the registration helper can take it automatically from there

MangoFVTT (approved):
LGTM! There's definitely improvements to be made but this is a great first step and I don't see any huge issues with the logic so far, well done 👍

src/system/utils/handlebars/index.ts
stanavdb: Yeah let's merge chat cards in first, then I'll fix the merge issues on this PR

zithith:
This is great! Much more than a first pas ;) And I can't see any issues with the code that you've written. Really well executed.

I am curious about some structuring decisions though...
* Firstly, was there not some way we could have leaned on existing libraries some how? e.g. D2/D3, ECharts, Google Charts, Mermaid? I know it wouldn't be quite as simple as plug and play, we'd need to figure out how to integrate the drag-and-drop, plus make sure they are style-able enough, but would it have saved a fair number of lines and custom component headache?
* We seem to be looping through the collection a lot. Why aren't we grabbing descriptions, context, connections etc when the talent is added, then doing minimal functions (e.g. HTML Enriching) during prepare context? Also, could a lot of these be stored inside the ExtendedNode object in the first place? (I appreciate it may be worth keeping separate maps if it means we aren't adding unnecessary entries into the DataModel schema). I may be missing something key in the foundry item/sheet lifecycle though, so I just want to be sure. It could also be only minimal impact, I haven't loaded up a copy of the branch locally to see.

Good luck with the next phase!

stanavdb:
>     * Firstly, was there not some way we could have leaned on existing libraries some how? e.g. D2/D3, ECharts, Google Charts, Mermaid? I know it wouldn't be quite as simple as plug and play, we'd need to figure out how to integrate the drag-and-drop, plus make sure they are style-able enough, but would it have saved a fair number of lines and custom component headache?

The rendering of the tree/connections wasn't actually all that complex to set up. I figured if I'd integrated some library I'd then have to figure out how to integrate that with drag-drop (specifically making nodes draggable), which seemed like a nightmare.

> We seem to be looping through the collection a lot. Why aren't we grabbing descriptions, context, connections etc when the talent is added, then doing minimal functions (e.g. HTML Enriching) during prepare context? Also, could a lot of these be stored inside the ExtendedNode object in the first place? (I appreciate it may be worth keeping separate maps if it means we aren't adding unnecessary entries into the DataModel schema). I may be missing something key in the foundry item/sheet lifecycle though, so I just want to be sure. It could also be only minimal impact, I haven't loaded up a copy of the branch locally to see.

We could probably fold `prepareNodeDescriptions` in with `extendNodeData` but we'll always need to do it on `prepareContext` anyway as the items in the nodes can be updated independently of the talent tree.  The items aren't actually embedded in the tree (items can't embed other items), they're just referenced by uuid. Same goes for connections, they're derived from the prerequisites and those can be changed between renders.

In the end, we'll likely end up refactoring talent trees when we get to leveling so I'm not too worried about it.

PR #171, opened 2024-11-21
feat(chat cards): add damage criticals and styling

### LINKED ISSUE #152: Support Damage Roll Criticals + Styling for Criticals
Damage rolls currently have no understanding of "critical" and no framework in for place for becoming critical rolls (i.e. maximising all dice). There is no different indication in the chat card styling for when a damage roll is a crit either.

We can tap into the Roll.maximise function quite easily to achieve this. All that needs to be setup are some additional properties to the DamageRoll class, and the relevant HTML enrichers in the chat message to allow for upgrading the roll to a crit, and to properly display styling for that crit.

**Checklist:**
- [X] I have searched the existing issues to ensure this feature has not already been requested.

MangoFVTT:
**Type**  

- [ ] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  
Implements all the functionality related to damage rolls having criticals and being able to be upgraded to criticals via the chat card overlays. Upgrading to a critical essentially just maximizes all dice.

**Related Issue**  
Closes #152 

**How Has This Been Tested?**  
Tested on all cards that have damage.

image

**Checklist:**  


stanavdb (dismissed):

MangoFVTT:
> This is pretty much spot on to how I was thinking about it 👍🏻 My only concern now is that we seem to lose the original roll. I wonder if there's any issues with us rolling the critical pool silently as we do with the graze and then just shifting between showing one or the other when the toggle button is clicked? Could be another iteration though

Once a roll is upgraded to a crit there's no reason to keep the old roll no? It's the same with adding advantage to the d20s or whatever, the roll is irrevocably changed. I think that's fine because people aren't going to be randomly clicking the upgrade to crit unless they explicitly want a crit, presumably

zithith:
> Once a roll is upgraded to a crit there's no reason to keep the old roll no? It's the same with adding advantage to the d20s or whatever, the roll is irrevocably changed. I think that's fine because people aren't going to be randomly clicking the upgrade to crit unless they explicitly want a crit, presumably

I'm imagining a scenario where a player sees the opportunity and then decides to crit almost without thinking and then someone else at the table points out that they could use the opportunity in a more advantageous/interesting way. Think we'd want some way to take it back. Or for the sake of mis-clicks

MangoFVTT:
> > Once a roll is upgraded to a crit there's no reason to keep the old roll no? It's the same with adding advantage to the d20s or whatever, the roll is irrevocably changed. I think that's fine because people aren't going to be randomly clicking the upgrade to crit unless they explicitly want a crit, presumably
> 
> I'm imagining a scenario where a player sees the opportunity and then decides to crit almost without thinking and then someone else at the table points out that they could use the opportunity in a more advantageous/interesting way. Think we'd want some way to take it back. Or for the sake of mis-clicks

My problem with this is that we introduce an expectation of backtracking rolls, which we would then need to apply to every part of the roll. This will in turn make our lives moving forward extremely annoying. Off the top of my head, if we want to allow backtracking crits, we should also allow the following:
- backtracking the d20 overlay buttons
- backtracking any advantage/disadvantage selected accidentally in the roll config
- backtracking whatever system we put into place later to allow advantage/disadvantage on individual dice

This is going to make the chat card logic extremely convoluted and bloated, because we will need to essentially track a history state for everything. And how far do we take that? Let's say we put into place later that users can click individual dice in tooltips to apply advantage/disadvantage to them, do we only allow backtracking once? Do we allow backtracking indefinitely until the original state of the roll? How much history do we want to store and at what complexity level? On top of that you have to also consider the UI becoming a bloated nightmare of options for people to choose.

In my opinion, we should make it difficult to make mistakes through good UI and good UX, but not try to cover every single thing that can go wrong. Sometimes imo it's ok to say: "hey you changed this roll irrevocably. if you want to go back you have to just make a different roll".

stanavdb:
I'm with Mango on this one. While I think the concern is valid, let's stay away from the complexity of dealing with action histories.

zithith:
I agree, we don't want a bloated UI, we would rather make mistakes harder to make by good UI. Sure. But I think we're just as fair to say "we're only allowing take-backs in these specific circumstances", cos it seems like this would be an easy win. And I see the issue of changing your mind on an opportunity spend to be a common one that GMs would allow (having played my fair share of Star Wars/Genesys). Back tracking advantage isn't so necessary as you can still see both rolls so you don't have to reroll to manually ignore it, but here we're getting rid of the previous info.

That said, I'm quite happy to merge the branch and re-address the toggling later on.

stanavdb (approved):

zithith (approved):

PR #174, opened 2024-11-21
feat(chat cards): add styling for damage notifications

MangoFVTT:
**Type**  

- [ ] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  
Styles the damage notification messages and moves the enrichment behaviour to the chat message for centralisation. Also fixes the undo damage/healing button that was broken by prior chat styling.

Also, I fixed the deflect calculations so that deflect carries across multiple damage types if necessary (instead of applying deflect to every type individually). This fix also makes the total deflected be accurately represented in the final damage calculation shown in the message, which it wasn't previously if the total deflected was less than the maximum deflect value.

**How Has This Been Tested?**  
Tested with both healing and damage applied, across both characters and adversaries.

image

**Checklist:**  


src/system/documents/actor.ts
```diff
@@ -565,71 +565,57 @@ export class CosmereActor<
...
             }
 
             if (instance.type === DamageType.Healing) {
-                amount = -amount;
+                amount *= -1;
             }
 
             // Add to running total
             damage += amount;
         });
```
stanavdb, line 586: Not particularly important point but since we no longer have to apply deflect separately we can probably just sum up all deflectable damage and reduce it by the deflect in one go. Should make this a bit more readable.

src/system/documents/chat-message.ts
```diff
@@ -17,6 +17,7 @@ export const MESSAGE_TYPES = {
     SKILL: 'skill',
     ACTION: 'action',
     INJURY: 'injury',
+    TAKEN: 'taken',
```
stanavdb, line 20: Can we rename this to something more clear? I appreciate the shorthand but `taken` feels a bit too abstract.

stanavdb (commented):
Looks good to me! Just two minor points.

zithith (dismissed):
I agree with Stan's comments about readability, but happy enough with what it's doing as is 👍🏻

stanavdb (changes requested):
Code looks good but I ran into two things when testing: 
1. It's not super clear that the damage has been undone. Perhaps we can strikethrough the damage text?
image
2. Users that do not have permission to modify the actor can still see and click the undo option. (Which then obviously causes an error)

stanavdb (approved):

zithith (approved):

PR #175, opened 2024-11-21
Fix for 157 - Unarmed graze damage

### LINKED ISSUE #157: Unarmed Stike Fails to Roll
**Describe the bug**  
UnarmStrike Fails to roll or output to chat.

**To Reproduce**  
1. _Go to character sheet with an unarmed attack weapon
2. _Click on the roll for the unarmed attack
3. _See that it does not output

**Expected behavior**  
Unarmed attack should roll (even if it just uses flat damage and not the scaling damage)

**Screenshots**  
image

**Foundry VTT Version**  
12.331

0.2.0

**Browser**  
Chrome

**Additional context**  
Tried with a new PC and a already created on from the last update. Same issue

**Checklist:**

zithith:
**Type**  
- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
Updated the graze formula fall back for when the base damage formula doesn't include dice.

**Related Issue**  
Closes #157 

**How Has This Been Tested?**  
Locally rolling unarmed attacks and standard attacks

**Checklist:**  
- [ ] My changes do not introduce any new warnings or errors.
- [ ] My PR does not contain any copyrighted works that I do not have permission to use.

**Additional context**  


zithith:
This isn't a full fix, see code comments

src/system/documents/chat-message.ts
```diff
@@ -261,8 +261,7 @@ export class CosmereChatMessage extends ChatMessage {
             const tooltipNormal = $(await rollNormal.getTooltip());
             this.enrichDamageTooltip(rollNormal, type, tooltipNormal);
             tooltipNormalHTML +=
-                tooltipNormal.find('.tooltip-part')[0].outerHTML;
-
+                tooltipNormal.find('.tooltip-part')[0]?.outerHTML || ``;
```
zithith, line 264: this empty string alternative is a clutch.
It works, but the tooltips are no longer consistent. Screenshot:
image

MangoFVTT: Hmm this is annoying yeah. The problem is that the core foundry code only generates tooltips for dice. To show constants I do enrichment AFTER the tooltip is generated but that requires an already existing tooltip which obviously doesn't exist here. Not sure what the best fix would be unfortunately, i'd have to have a proper think about even more complicated tooltip generation code. I'll take a look at what the dnd5e system does in this case to get ideas

src/system/documents/item.ts
```diff
@@ -518,9 +518,14 @@ export class CosmereItem<
         );
 
         // Make the graze pool and roll it
+        const defaultGraze = roll.terms.every(
+            (term) => !(term instanceof foundry.dice.terms.DiceTerm),
+        )
```
MangoFVTT, line 523: Minor nitpick: it's faster to do `!roll.terms.some(term => foundry.dice.terms.DiceTerm)` because that will exit out early if a single DiceTerm is found, whereas every will always run the full iteration

stanavdb:
My commit fixes the tooltip issue. Did however run into a different (not quite related) bug; If a damage roll has multiple different dice, the tooltip doesn't render those correctly, it just drops anything beyond the first die type. 
image
image

MangoFVTT:
> Did however run into a different (not quite related) bug; If a damage roll has multiple different dice, the tooltip doesn't render those correctly, it just drops anything beyond the first die type.

Interesting 🤔 the dice tooltip is generated by foundry itself and injected into a handlebars template, so it may be an error in the template that's causing the issue. I'll have to check when I have time

The rest looks good to me

zithith:
Great work Stan! Good solutions, I like everything here.
The multi dice thing can be raised as a separate issue?

stanavdb (approved):

PR #189, opened 2024-11-30
Feat/flat modifiers to attacks 135

### LINKED ISSUE #135: Add flat modifiers to attacks
Sometimes when setting up an attack want to use a flat number instead of a skill

Should be an alternative where you can just put in a number to you as the modifier instead of having to select a skill

**Additional context**  
Suggested by `@stanvdb` on 22/10/2024

**Checklist:**

zithith:
**Type**  
- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
This adds a new input box to the item activation form (for skill roll based activations) that allows users to input a formula and that will get added to the attack roll calculations. Using the existing roll data objects, there are a range of `@` tags for various modifiers and flat numbers and dice can be included to allow for manually representing global effects.

**Related Issue**  
`Closes #135`

**How Has This Been Tested?**  
Loaded build into my local server and tested using GM and Player accounts.
1. Loaded Bridge 9 module into server
2. For the global item modifiers:
    1. Opened the sheet of any of the pre-gens, switch to the inventory tab
    2. Edited a weapon to include a value in the new field
    3. Rolled the item
        - Confirmed the values show up in the formula display
        - Checked that any dice added appear in DSN animation
        - Confirmed that dice values and flat numbers are shown in the chat card and included in the totalling of the outputted value
    4. repeat ii and iii for various scenarios (dice included, use `@mod`/`@attribute`/`@skill.rank`, use `@attr`/`@skills.<skill>.ranks` for any value on the character)
3. For the situational bonuses:
    1. Open the sheet of a pre-gen
    2. Click on an activation icon for any interaction needing a d20 roll
    3. Enter a value in the Temporary Modifier field
        - Confirmed the formula display updated with the new value
    4. Made the Roll
        - Confirmed the roll still works
        - Confirmed DSN still works
        - Confirmed the values showed up in the chat card
        - Confirmed advantage/disadvantage not broken
        - Confirmed shortcut keys not broken
        - Confirmed Attacks don't double-dip the temporary mods
    5. Repeat for various values (+ve and -ve additions, dice, flat numerics, `@` notations, empty field) and for rolls from items as well as weapons and straight skill rolls

Global Mod:

||
|:-:|
|image|
|(I changed the label text since the screenshot, it is now "Roll Formula Extension")<br />⬇️|
|image|
|⬇️|
|image|
|⬇️|
|image|

Temp Mods:

||
|:-:|
|image|
|⬇️|
|image|
|⬇️|
|image|

**Checklist:**  
- [x] I have commented on my code, particularly in hard-to-understand areas. *Happy to add more*

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -861,7 +861,10 @@
                 "ReduceFocus": "Click to reduce focus from selected token(s).",
                 "RollCrit": "Click to roll critical damage.",
                 "RollAdvantage": "Click to roll with advantage.",
-                "RollDisadvantage": "Click to roll with disadvantage."
+                "RollDisadvantage": "Click to roll with disadvantage.",
+                "ApplyInjury": "Click to apply injury to character.",
+                "UndoDamage": "Click to undo damage taken.",
+                "UndoHealing": "Click to undo healing."
             },
             "Trays": {
                 "Targets": "Targets"
@@ -875,8 +878,6 @@
             "ApplyDamage": "{actor} takes <strong>{amount} damage</strong>.",
             "ApplyHealing": "{actor} recovers <strong>{amount} health</strong>.",
             "DamageCalculation": "Damage Calculation: {calculation}",
-            "UndoDamage": "Click to undo damage taken.",
-            "UndoHealing": "Click to undo healing.",
             "ViewRollsDetails": "Roll Details",
             "Welcome": "<h2>Welcome to the Cosmere RPG System!</h2> <p>Thank you for trying out the community-developed Cosmere Roleplaying Game system! You're currently using version <strong>{version}</strong>. As this is an early release, some bugs or missing features are to be expected, but your feedback will help us improve the system. </p> <p>For discussions, updates, and support, join us on the <a href=\"{discordLink}\">Metalworks Discord</a> server.</p> <p>If you encounter any issues or have suggestions, we'd love to hear from you! Please report them on our <a href=\"{issuesLink}\">GitHub Issues page</a> or on the <a href=\"{discordLink}\">Metalworks Discord</a>. Be sure to review the <a href=\"{contributingLink}\">contribution guidelines</a> if you'd like to get involved in development. </p> <p>We're excited to have you on board and hope you enjoy playing in the Cosmere!</p>"
         },
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -1010,6 +1010,7 @@
     "GENERIC": {
         "Unknown": "Unknown",
         "None": "None",
+        "Custom": "Custom",
         "Default": "Default",
         "Document": "Document",
         "Formula": "Formula",
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -678,7 +678,7 @@
                     "Consume": "Resource Consumption",
                     "Uses": "Limited Uses",
                     "Recharge": "Recharge",
-                    "AdditionalFormula": "Attack Formula Extension",
+                    "AdditionalFormula": "Roll Formula Extension",
                     "AdditionalFormulaDescription": "Input any extra modifiers you wish to add to this item's d20 Skill Roll. This input will be treated as a dice formula, but any die you add won't be available for advantage/disadvantage currently. Paramters you can use: @mod, @attribute & @skill.rank refer to the values of the skills selected here, but the whole list of skills and attributes can be accessed - ask on the discord for help!"
                 },
                 "Attack": {
@@ -906,7 +906,8 @@
             "Apply": "Apply",
             "Graze": "Graze",
             "Full": "Full"
-        }
+        },
+        "TemporaryBonus": "Temporary Modifiers (formula)"
     },
     "ROLLS": {
         "Recovery": "[character] rolls their recovery die.",
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -677,7 +677,9 @@
                     "Cost": "Activation Cost",
                     "Consume": "Resource Consumption",
                     "Uses": "Limited Uses",
-                    "Recharge": "Recharge"
+                    "Recharge": "Recharge",
+                    "AdditionalFormula": "Attack Formula Extension",
+                    "AdditionalFormulaDescription": "Input any extra modifiers you wish to add to this item's d20 Skill Roll. This input will be treated as a dice formula, but any die you add won't be available for advantage/disadvantage currently. Paramters you can use: @mod, @attribute & @skill.rank refer to the values of the skills selected here, but the whole list of skills and attributes can be accessed - ask on the discord for help!"
                 },
                 "Attack": {
                     "Title": "Attack",
--- a/src/system/applications/dialogs/attack-configuration.ts
+++ b/src/system/applications/dialogs/attack-configuration.ts
@@ -1,8 +1,12 @@
 import { Attribute } from '@system/types/cosmere';
 import { RollMode } from '@system/dice/types';
 import { AdvantageMode } from '@system/types/roll';
-import { AnyObject, NONE, Noneable } from '@system/types/utils';
-import { isNone } from '@src/system/utils/generic';
+import { AnyObject, NONE, Nullable } from '@system/types/utils';
+import {
+    getFormulaDisplayString,
+    getNullableFromFormInput,
+    isNull,
+} from '@src/system/utils/generic';
 
 import { D20RollData } from '@system/dice/d20-roll';
 import { DamageRollData } from '@system/dice/damage-roll';
@@ -84,7 +88,7 @@ export namespace AttackConfigurationDialog {
         /**
          * The attribute that is used for the roll by default
          */
-        defaultAttribute?: Noneable<Attribute>;
+        defaultAttribute?: Nullable<Attribute>;
 
         /**
          * The roll mode that should be selected by default
@@ -93,7 +97,7 @@ export namespace AttackConfigurationDialog {
     }
 
     export interface Result {
-        attribute: Noneable<Attribute>;
+        attribute: Nullable<Attribute>;
         rollMode: RollMode;
         skillTest: {
             plotDie: boolean;
@@ -193,7 +197,9 @@ export class AttackConfigurationDialog extends ComponentHandlebarsApplicationMix
     ) {
         if (event instanceof SubmitEvent) return;
 
-        const attribute = formData.get('attribute') as Noneable<Attribute>;
+        const attribute = getNullableFromFormInput<Attribute>(
+            formData.get('attribute') as string,
+        );
         const rollMode = formData.get('rollMode') as RollMode;
         const plotDie = formData.get('plotDie') === 'true';
         const tempMod = formData.get('temporaryMod')?.valueOf() as string;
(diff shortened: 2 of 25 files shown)


src/system/applications/item/base.ts
```diff
@@ -116,6 +116,17 @@ export class BaseItemSheet extends TabsApplicationMixin(
...
                 formData.set('system.activation.skill', null);
 
+            if (
+                'system.activation.attribute' in formData.object &&
+                formData.object['system.activation.attribute'] === 'default'
+            ) {
+                formData.set(
+                    'system.activation.attribute',
+                    CONFIG.COSMERE.skills[
+                        formData.object['system.activation.skill'] as Skill
```
zithith, line 126: Note: This wasn't possible when the skills were defined as enums

zithith:
Also, apologies about pulling in the workflow changes from main. got my target branches a bit confused at one point before I realised I had updated the feature branch. Hopefully we can cherry pick those back out when squashing or just let it lie, as it would need to have been pulled before a release anyhow?

src/lang/en.json
```diff
@@ -677,7 +677,9 @@
                     "Cost": "Activation Cost",
                     "Consume": "Resource Consumption",
                     "Uses": "Limited Uses",
-                    "Recharge": "Recharge"
+                    "Recharge": "Recharge",
+                    "AdditionalFormula": "Roll Formula Extension",
+                    "AdditionalFormulaDescription": "Input any extra modifiers you wish to add to this item's d20 Skill Roll. This input will be treated as a dice formula, but any die you add won't be available for advantage/disadvantage currently. Paramters you can use: @mod, @attribute & @skill.rank refer to the values of the skills selected here, but the whole list of skills and attributes can be accessed - ask on the discord for help!"
                 },
```
MangoFVTT, line 683: I think it's preferrable for this to be an on-hover tooltip rather than a description that's always visible. It buys us some real estate in the sheeet

src/system/applications/dialogs/roll-configuration.ts
```diff
@@ -29,6 +29,11 @@ export namespace RollConfigurationDialog {
          */
         parts: string[];
 
+        /**
+         * A dice formula stating any miscellanious other bonuses or negatives to the specific roll
```
MangoFVTT, line 33: miscellaneous*

src/system/config.ts
```diff
@@ -8,7 +8,7 @@ import {
     Attribute,
     AttributeGroup,
     Resource,
-    Skill,
+    Skills,
```
MangoFVTT, line 11: Is there any reason to make this plural? All the other config enums are singular and that is common naming practice for enums

MangoFVTT (changes requested):

src/system/config.ts
zithith: Because it's no longer an enum. It's now the list of skills as a constant.
Which in effect makes it like an enum, but means you can do more with how you access it. i.e. in this PR I had to use the string value passed in from a select box to select the right value, only as enum entries are treated kinda like Symbols they very much can't be swapped for their string equivalents. This way you can do that.

Then you use a little TS trick to get the types of the Keys as a union type which can act in places where you would have said "expect an entry from this enum", and *that* type is singular because it means this will be a singular instance of one of these constants.

src/system/applications/dialogs/roll-configuration.ts
zithith: Damn, thanks. Don't know why Code Spell Check didn't pick up on that. Annoying as it was late and I felt like the word wasn't right, but wasn't able to see why 🤣

src/lang/en.json
zithith: I don't disagree, I just don't know if we use tooltips for providing info like that elsewhere in any of the sheets yet, do we?
This is where something like a set of the tutorials/walkthroughs ( or a wiki

MangoFVTT: I'm all for moving any hint infos to tooltips wherever we can. It's just cleaner UI with less overwhelming amounts of text

zithith:
Regarding the "Noneable" stuff, feel free to suggest another name. I wanted to use Nullable but that feels like it's inaccurate and also likely to clash with potential future 3rd party libraries.
If you guys are happy with it, I'll raise another issue/PR to go through and standardise the use of "none" where appropriate.

src/system/applications/dialogs/attack-configuration.ts
```diff
@@ -269,15 +291,21 @@ export class AttackConfigurationDialog extends ComponentHandlebarsApplicationMix
 
     protected _prepareContext() {
         const skillTestFormula = foundry.dice.Roll.replaceFormulaData(
-            this.data.skillTest.parts.join(' + '),
+            this.data.skillTest.parts
+                .join(' + ')
+                .replace(/\+ -/g, '-')
+                .replace(/\+ \+/g, '+'),
```
zithith, line 297: this should probably become it's own function (getFormulaString or something)

src/system/applications/item/base.ts
stanavdb: This works just fine with the skills defined as an enum? Whether the skills are an enum or not shouldn't have any impact here, as you're just casting the type; the generated js will be identical.

I pulled in the branch and converted skills back to an enum to test and the build works without any further changes.

stanavdb (commented):
Is there a reason to step away from using `null` to represent none selections and move to explicit `'none'`?

`Noneable` is useful within the applications that make the selection. But beyond that context I feel `null` is a cleaner solution.

stanavdb (commented):
Setting the skill to `None` for any skill test (or attack), causes the activation to no longer work as both the `rollAttack` and the `roll` function on the item document explicitly check for skill.

zithith:
> Setting the skill to `None` for any skill test (or attack), causes the activation to no longer work as both the `rollAttack` and the `roll` function on the item document explicitly check for skill.

Ok, I'll go back and check that, been focussing on the attribute for the last few sessions. TBF I don't think I changed any of the logic for skill selection as it seemed like that was already handled. Might just have to go in and add belt and braces around it.

zithith:
> Is there a reason to step away from using `null` to represent none selections and move to explicit `'none'`?
> 
> `Noneable` is useful within the applications that make the selection. But beyond that context I feel `null` is a cleaner solution.

That was partly my reason for mentioning/bringing attention to it in my earlier comments.
I don't know for sure. Mostly I was following the approach that the Skills and other dropdowns already had in place for selecting a "No Value" option.

The model value is still set to null (see base item line 134). This is mostly for handling where the select option value is handed around the application.

There is probably some way that this could be converted to a null, but technically speaking I'm not sure if using explicit null is good practice in TS and that having a separate value to identify when the user has actively selected a No Value might be cleaner. I'll have to check some articles/ask my work colleagues.

src/system/applications/item/base.ts
zithith: Interesting... I was getting error highlights in the IDE, and confirmed my understanding with a Google that you can't index into an enum with a string and you can't convert the enum key into string either.

I'm away from home for a few days, but will have to try reverting this when I get back.

zithith: After double checking, I found that just casting the form output to an enum is doable... (odd how it gave me errors before that matched my understanding and reading on the topic 🤔).
So I've reverted the skills for consistency

src/system/config.ts
zithith: redundant now as the enum change is now reverted

src/system/applications/dialogs/attack-configuration.ts
zithith: done

src/system/dice/index.ts
```diff
@@ -89,7 +89,11 @@ export async function d20Roll(
 
     // Construct the roll
     const roll = new D20Roll(
-        ['1d20'].concat(config.parts ?? []).join(' + '),
+        ['1d20']
+            .concat(config.parts ?? [])
+            .join(' + ')
+            .replace(/\+ -/g, '-')
+            .replace(/\+ \+/g, '+'),
```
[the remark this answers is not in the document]
zithith: done

zithith:
OK, so I've gone in and updated to what could be the final approach?
I see why the rolling was failing on no-skill. We were short-circuiting the roll functions and using the "none" option as a sort of "this is unconfigured" flag. Removed as that would invalidate some of the idea for this feature.

Enabling no-skill rolls highlighted that the use of "none" as a string value was causing some problems buried deep in some of the object assignments so instead of trying to go round the houses I've switched it to rely on nulls from the earliest opportunity. I think there could be some discussion had to align our use of null vs. undefined in places, but for now happy to take this route.

Word of note: Would be good to get more testing in. There seems to be a bit of flakiness to when the rolls pick up `@` variables properly in the current build. flat numbers seem to work across the board for global mods and for temporary ones, but the variables seem to sometimes read as 0 in the calculations despite reading their correct values in the config dialog.

src/system/applications/dialogs/attack-configuration.ts
```diff
@@ -207,7 +213,7 @@ export class AttackConfigurationDialog extends ComponentHandlebarsApplicationMix
         this.data.skillTest.temporaryModifiers = tempMod;
 
         const skill = this.data.skillTest.data.skill;
-        const attributeData = !isNone(attribute)
+        const attributeData = !isNull(attribute)
```
stanavdb, line 216: Do we want to keep the `isNull` check in? One of the advantages of using `null` over `'none'` is that it's a falsy value

src/system/applications/dialogs/attack-configuration.ts
```diff
@@ -216,7 +222,7 @@ export class AttackConfigurationDialog extends ComponentHandlebarsApplicationMix
         this.data.skillTest.data.mod = rank + value;
         this.data.skillTest.plotDie = plotDie;
 
-        this.data.defaultAttribute = !isNone(attribute) ? attribute : undefined;
+        this.data.defaultAttribute = !isNull(attribute) ? attribute : undefined;
```
stanavdb, line 225: Same question here, but instead we can also use the nullish coalescing operator for this.

src/system/documents/item.ts
```diff
@@ -398,7 +399,7 @@ export class CosmereItem<
 
         // Set up actor data
         const data: D20RollData = this.getSkillTestRollData(
-            skillId,
+            skillId ? skillId : null,
```
stanavdb, line 402: nullish coalescing would be neater here

src/system/types/utils.ts
```diff
@@ -4,10 +4,11 @@ export {
     EmptyObject,
 } from '@league-of-foundry-developers/foundry-vtt-types/src/types/utils.mjs';
 
+// Constant to improve UI consistency
 export const NONE = 'none';
-export type None = typeof NONE;
 
-export type Noneable<T> = T | None;
+// Simple utility type for easier null definitions, but general rule: only use it when you have one type that is nullable (i.e. prefer X | Y | null over Nullable<X | Y>)
+export type Nullable<T> = T | null;
```
stanavdb, line 11: Is there a reason to use this over `Type | null`? Not opposed to this, more questioning for consistency

stanavdb (commented):
Left a couple of small comments. Haven't tested yet but looks like we're pretty much there.

src/system/applications/dialogs/attack-configuration.ts
zithith: True, I just didn't want to keep typing out the check long hand, so renamed the existing function once 😂
I can go back and swap it out next session I sit down at VScode

src/system/types/utils.ts
zithith: None technically, I did look it up.

src/lang/en.json
zithith: I think we can leave this until your style changes are in and the work is done in the item cards too.
Then we can take a look at any hint text across the system and come up with a standardised pattern for displaying that data/including tooltips

MangoFVTT (approved):
Seems good to me, any extra styling stuff we need to do will get rolled into the 0.3.0 changes 👍

PR #195, opened 2024-12-14
Inline Tabs for Actor sheets - Pop-Out! workaround

### LINKED ISSUE #194: Enable support for Pop-out module (workaround)
When using the Pop-Out! module, the side tabs approach doesn't work.

Add support for pop-out (make the various tabs useable while the sheets are not in the FoundryVTT window). This will be a quick workaround for those using the module for now: add in a toggle setting for side tabs on the character sheets as we have for the item sheets.

A fix that preserves the side tabs may be possible but will need more investigation.

**Version**  
0.2.1

**Checklist:**

zithith:
**Type**  

- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
Added a setting toggle (as per the item sheet one, but inverted) to make the tabs for actor sheets fall in-line, more like the default foundry behaviour. Required a small re-factor of the navigation parts for the actor sheets so that they appear in a relevant place in the sheet structure.

**Related Issue**  
Closes #194 

**How Has This Been Tested?**  
Local machine. Opened sheets with and without the setting toggled to see the change (doesn't re-render already opened sheets).

image
image

**Checklist:**  

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -1099,6 +1099,10 @@
         }
     },
     "SETTINGS": {
+        "actorSheetInlineTabs": {
+            "name": "Horizontal Inline Tabs for Actor Sheets",
+            "hint": "If enabled, actor sheets will use horizontal tabs within the bounds of the sheet, as per foundry default sheets. This setting is meant as a work-around solution for the Pop-Out! module, for those who want to use it."
+        },
         "itemSheetSideTabs": {
             "name": "Vertical Side Tabs for Item Sheets",
             "hint": "If enabled, item sheets use vertical tabs down the right-hand side, similar to the character sheet, instead of the default in-line horizontal ones."
--- a/src/style/sheets/actor/module.scss
+++ b/src/style/sheets/actor/module.scss
@@ -2,10 +2,37 @@
 @import './adversary.scss';
 
 .sheet.actor {
-    overflow: visible;
 
-    .window-content {
+    &.side-tabs {
         overflow: visible;
+
+        .window-content {
+            overflow: visible;
+        }
+
+        nav {
+            border: none;
+            flex-direction: column;
+            position: absolute;
+            left: 100%;
+            top: 15%;
+
+            > a {
+                background: rgba(11, 10, 19, 0.9);
+                width: 2.88867rem;
+                height: 2.5rem;
+                display: flex;
+                align-items: center;
+                justify-content: center;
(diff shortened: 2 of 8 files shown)


zithith:
Note the styling is not great... at all...
A) It's a workaround
B) I didn't want to upset the merge conflicts for mango's styling branches more than I had to.

I did make the decision that I left out the labels on the tabs as there are too many of them for the space on the sheet. Though if peopele think adding them back for accessibility is better, we could look at stacking icon over text?

src/lang/en.json
```diff
@@ -1099,6 +1099,10 @@
         }
     },
     "SETTINGS": {
+        "actorSheetInlineTabs": {
+            "name": "Horizontal Inline Tabs for Actor Sheets",
+            "hint": "If enabled, actor sheets will use horizontal tabs within the bounds of the sheet, as per foundry default sheets. This setting is meant as a work-around solution for the Pop-Out! module, for those who want to use it."
+        },
         "itemSheetSideTabs": {
             "name": "Vertical Side Tabs for Item Sheets",
             "hint": "If enabled, item sheets use vertical tabs down the right-hand side, similar to the character sheet, instead of the default in-line horizontal ones."
```
MangoFVTT, line 1108: I think we should make the setting consistent here where on = vertical  and off = horizontal. It doesn't make sense for the 2 settings to do opposite things

MangoFVTT (approved):
This is a fine change for 0.2.2, but will need to be discarded and rebuilt for 0.3.0 I suspect, since the CSS layout won't translate properly

src/lang/en.json
zithith: I had considered that, but I felt it made more sense to me that the defaults for any optional settings are "off" and that those defaults set it to the way we are suggesting people start.

zithith:
> This is a fine change for 0.2.2, but will need to be discarded and rebuilt for 0.3.0 I suspect, since the CSS layout won't translate properly

Yeah, looking at the handlebars changes for your styling PR as well this seems like it would be almost entirely re-worked. Probably best to just strike it from 0.2.2 so we don't end up reverting functionality, especially as previously mentioned, it doesn't seem to do what people were after (at least according to my testing).

zithith:
Closing PR as changes are somewhat redundant, will leave the issue to be investigated some more in the 0.3.0 - 1.0 development window

Issue #70, opened 2024-10-06
Level-up / Advancement / Character Gen


zithith:
**Is your feature request related to a problem? Please describe.**  
As the foundry implementation will be independent of Demiplane we want to provide the players a way to manage characters as well as play them.

**Describe the solution you'd like**  
Some form of process (probably a dialog) that supports users through selecting choices they are allowed at any point in their character progression.
GM grants/homebrew overrides to the core rules would be nice where possible, but probably a feature for second or third pass.

**Describe alternatives you've considered**  
Potentially we could look at doing something like a Demiplane importer, but that may not be possible and when release arrives it will ikely require the user to have made a demiplane/roll20 purchase too which seems counter-productive. An importer would still be a good feature to add on top of a built-in progression system later down the line but I feel it won't satisfy enough to replace it.

**Version**  
0.1.0

**Additional context**  
_None_

**Checklist:**
- [x] I have searched the existing issues to ensure this feature has not already been requested.

stanavdb:
@zithith We should probably split this up into a couple of issues as well. Currently this encompasses:

- [x] Basic advancement (ability to set level & have health etc. scale)
- [x] Talent trees to be able to pick talents with prerequisites
- [ ] Character builder

zithith:
> @zithith We should probably split this up into a couple of issues as well. Currently this encompasses...

Yeah, that absolutely crossed my mind last night, just wanted to get your take on how to split it

zithith:
created a separate issue for the character builder to go in 1.0: #309

PR #200, opened 2025-01-16
feat(styling): sheet styling refactor and light mode theme

MangoFVTT:
**Type**  
What type of pull request is this? (e.g., Bug fix, Feature, Refactor, etc.)

- [ ] Bug fix
- [X] Feature
- [X] Refactor
- [ ] Other (please describe):

**Description**  
This is a giant PR that that completely refactors actor sheet CSS and fixes a host of issues with theming. In summary, this includes the following changes:
- Completely rebuild character and adversary sheets with new HTML and CSS classes, trying to reduce as much as possible duplicate styling in the CSS files. Made improvements as well to layout so that the sheet is usable in edit mode without any features getting hidden (e.g. injury rolling).
- Add all BW font files and change the application to use those. Remove the old font files. This also means changes to chat cards since they use these new fonts as well.
- Fix theme swapping and implement colours for both light/dark mode in both default/Stormlight themes.
- Fix a range of issues that came up with sheet rebuilding, and also changed various parts of sheet functionality to prepare for future features.

**Related Issue**  
Closes #188, #47, #44.

**How Has This Been Tested?**  
I've tested this and fixed a bunch of bugs already. There are more, for sure, since this is a major sheet refactor. I will continue to fix those bugs along the way as I proceed through styling.

**Screenshots (if applicable)**  
image
image
image
image

**Checklist:**  
- [X] I have commented on my code, particularly in hard-to-understand areas.
- [X] My changes do not introduce any new warnings or errors.
- [X] My PR does not contain any copyrighted works that I do not have permission to use.
- [X] I have tested my changes on Foundry VTT version: 12.331.

**Additional context**  
This is Part 1 of a major styling effort as we have discussed. I suspect this is the bulk of the work re: colouring and a lot of styling partials that can be re-used elsewhere, but there is plenty more work to come.

PR #202, opened 2025-01-20
feat(advancement): add basic advancement

stanavdb:
**Type**  

- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
This PR adds the basics of advancement: Level is editable through an input, health & max skill ranks are derived from level/advancement rules. Other than including the value in the character data model `maxSkillRanks` is currently unused.

**Related Issue**  
Partially addresses #70 

**How Has This Been Tested?**  
1. Open character
2. Toggle edit more
3. Update level
4. Check if health has increased (in accordance with advancement table)

image

**Checklist:**  


src/system/utils/advancement.ts
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Returns the all advancement rules between the start level (exclusive) and the end level (inclusive).
+ * If the end level is higher than the last rule, the last rule is repeated.
+ */
+export function getAdvancementRulesForLevelChange(
+    startLevel: number,
+    endLevel: number,
+): (AdvancementRuleConfig & { level: number })[] {
+    // Swap the levels if the end level is lower than the start level
```
zithith, line 36: I guess this is to allow for undo-ing/down-levelling type actions?

src/system/utils/advancement.ts
```diff
@@ -0,0 +1,163 @@
...
+        : getAdvancementRulesUpToLevel(levelOrRules);
+
+    // Calculate the health
+    return rules.reduce(
+        (health, rule) =>
+            health +
+            (rule.health ?? 0) +
+            (rule.healthIncludeStrength ? strength : 0),
+        0,
+    );
```
zithith, line 81: May be worth clarifying, but I don't think the increasing strength retroactively increases your health as this would do.
Rules say "increase your health" rather than "re-calculate" during the section on "gaining a level". Though with the clear 5e heritage they are leaning on, the retroactive bonus addition may be assumed as per what D&D does with Con...

But if that is the case there's no way to do this easily. Either we track the character history or we only ever adjust the maximum during a level change effect (which would be rather brittle I think)...

zithith (commented):
Good start.
I like some of the clever array magic you've done and keeping track of the grants for each level probably gives us a bit more control I think, but part of me wonders if we only need to keep the relevant tiers in the advancement rule config? And a lot of the stats can be derived with math formula rather than loops.
My approach still wouldn't be easily resilient to the ability bonus issue I mention in my other comment mind

stanavdb:
> Good start. I like some of the clever array magic you've done and keeping track of the grants for each level probably gives us a bit more control I think, but part of me wonders if we only need to keep the relevant tiers in the advancement rule config? And a lot of the stats can be derived with math formula rather than loops. My approach still wouldn't be easily resilient to the ability bonus issue I mention in my other comment mind

The reason I've gone for an array with each of the grants per level is that it allows for full configuration. I'd originally set it up to use formulas, however that locks advancement down on a system level which is awkward for homebrew and ultimately gives the same result.

src/system/utils/advancement.ts
stanavdb: Yep we should verify. I'm really hoping it does, otherwise this gets a lot more complex

src/system/utils/advancement.ts
stanavdb: Essentially, these utils came over from the full character builder prototype which needed to know what was lost if the level ever was lowered.

zithith: Yeah, that's fair. I don't have an issue with it being here. might as well stay, as I think we'll definitely need that later.
Maybe add a comment for others for now?

zithith:
> The reason I've gone for an array with each of the grants per level is that it allows for full configuration. I'd originally set it up to use formulas, however that locks advancement down on a system level which is awkward for homebrew and ultimately gives the same result.

I can accept that. I'm not sure how many things will want to mess around with core advancement like this (outside of a very distant chance of some 3rd Party release for Plotweaver) but my main issue was the amount of manual maintenance it might need, however it's not an ungodly amount more than the minimum required and it's unlikely to change much from a system PoV, so if we can get a bit more flexibility out of it I'm for it.

src/templates/actors/character/parts/header.hbs
```diff
@@ -92,7 +92,17 @@
...
+                {{#if isEditMode}}
+                    <input type="number"
+                        name="system.level",
+                        value="{{actor.system.level}}"
+                        min="1"
+                        step="1"
+                    >
+                {{else}}
+                    <span class="value">{{actor.system.level}}</span>
+                {{/if}}
```
MangoFVTT, line 105: I've been using input elements with a readonly tag (only applied if not in edit mode) instead of swapping between an input and a span in the 0.3.0 sheet refactor. It allows for more consistent styling without needing to style 2 different things. Also do note this needs to follow the new "sheet-stack" block format I've been using to align properly. You can take a look at how it is setup once that is merged

src/system/utils/advancement.ts
MangoFVTT: I second putting in a framework for undo-ing levels. When we get to more complete advancement later that auto-adds stuff like items and skills we definitely want a way to go back if a mistake is made (without manually tracking all the changes)

MangoFVTT (dismissed):
Good start, I would definitely look at the styling and templates post the #200 merge before considering merging this, we don't want anything to explode

src/templates/actors/character/parts/header.hbs
MangoFVTT: E.g. the document-name in the header:

```html
<h1 class="document-name">
     <input class="value" type="text" name="name" value="{{actor.name}}" {{#unless isEditMode}}readonly{{/unless}}> 
</h1>
```

zithith: > I've been using input elements with a readonly tag (only applied if not in edit mode) instead of swapping between an input and a span in the 0.3.0 sheet refactor. It allows for more consistent styling without needing to style 2 different things. Also do note this needs to follow the new "sheet-stack" block format I've been using to align properly. You can take a look at how it is setup once that is merged

The read-only approach is also more accessible for what it's worth

src/system/utils/advancement.ts
zithith: This was verified by DocSun, the specific rule is explained in the Strength description of the Attributes section; and this implementation is correct.

zithith (dismissed):

src/style/sheets/actor/module.scss
```diff
@@ -238,6 +238,17 @@
...
+
+                    input {
+                        position: absolute;
+                        left: 0;
+                        top: 0;
+                        width: 100%;
+                        height: 100%;
+                        padding: 5px 0 0 0;
+                        z-index: 1;
+                        text-align: center;
```
MangoFVTT, line 250: I think most of these are done already by the "value" class in the sheet-stack block. I could be wrong, but worth checking to avoid duplicate CSS (these files are long enough as is)

MangoFVTT (approved):

MangoFVTT (approved):

PR #216, opened 2025-02-03
feat(talent trees): refactor talent trees to use canvas

stanavdb:
**Type**  

- [ ] Bug fix
- [ ] Feature
- [x] Refactor

**Description**  
This PR refactors refactors the talent trees to be rendered by canvas rather than straight html and refactors the data model to move prerequisites over to the trees. This does not yet include support for nested trees, but the framework set up by this PR does allow for them.
Also adds support for some basic functionality that was missing from talent trees 

**Related Issue**  
Partially addresses #70 

**How Has This Been Tested?**  
Locally creating/modifying a talent tree.

**Checklist:**  


zithith:
> I'm not hugely familiar with Canvas so I can't for certain say everything here is perfect, but from a code standpoint it seems fine to me. I didn't really bother reviewing the styling much since it wil likely be refurbished later so as long as its functional for now (which it seems to be) then that is fine 👍

I'm very much in this same boat. I have half an idea of what the foundations of what I'm looking at should be, but lacking the detail on the maths without a solid sit down and there are so many APIs in here I don't know the first thing about.
I'm going to do one last scan for obvious pitfalls but then I'll rely on having seen Stan demo it to me on discord as proof the core functionality is stable, then we can iron out bugs as they turn up and more of us understand the ins and outs better.

zithith (approved):
aside from a couple of commented blocks that have been left, which I assume are there because this is still in-progress effectively, I'm good.

PR #219, opened 2025-02-08
fix(data models): fix issue causing derived statistics not to be updated when a bonus is applied

### LINKED ISSUE #217: Mod Changes to Attributres do not scale Movement Rate and Sense Range properly
**Describe the bug**  
One random thing I noticed when I did a modifier for AWA is that it did not change the actors sense range same with a move range with spd mod. which should dynamically change with modifiers.
The only things that SHOULDNT change with Attribute mods changing their base is Defenses (this seems to work correctly), max hp/focus(also good), and max invest (auto calc not in system yet so doesnt work). Honestly just seems like the default assumption is that mods only affect the skills going up which isnt true and should be adjusted.

**To Reproduce**  
1.create a item to add a mod to AWA or SPD stat 

Image

2. Make sure that the modifer would jump the Attribute up to the next tier of movement or sense range.
3. see that it does not work with sense or movement but that revelant skills related to the attribute do.

**Expected behavior**  
Sense or Movement range go up, as do all relevant skill modifiers. does not increase related defence, or max hp/focus/invest

**Foundry VTT Version**  
12.221

0.2.2

**Browser**  
Chrome

**Checklist:**

stanavdb:
*Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**
This PR makes it so the attribute bonus is correctly applied to derived statistics (e.g. movement rate)

**Related Issue**  
Closes #217 

**How Has This Been Tested?**  
1. Add effect that modifies an attribute bonus (`system.attributes.spd.bonus` to +2)
2. Toggle effect on
3. See movement rate increase

**Checklist:**  

### DIFF
--- a/src/system/data/actor/common.ts
+++ b/src/system/data/actor/common.ts
@@ -56,7 +56,7 @@ interface CurrencyDenominationData {
     convertedValue: Derived<number>;
 }
 
-interface AttributeData {
+export interface AttributeData {
     value: number;
     bonus: number;
 }
--- a/src/system/data/actor/common.ts
+++ b/src/system/data/actor/common.ts
@@ -56,6 +56,11 @@ interface CurrencyDenominationData {
     convertedValue: Derived<number>;
 }
 
+interface AttributeData {
+    value: number;
+    bonus: number;
+}
+
 export interface CommonActorData {
     size: Size;
     type: {
@@ -71,7 +76,7 @@ export interface CommonActorData {
         damage: DamageType[];
         condition: Condition[];
     };
-    attributes: Record<Attribute, { value: number; bonus: number }>;
+    attributes: Record<Attribute, AttributeData>;
     defenses: Record<AttributeGroup, { value: Derived<number>; bonus: number }>;
     deflect: DeflectData;
     resources: Record<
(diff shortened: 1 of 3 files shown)


src/system/data/actor/common.ts
```diff
@@ -56,6 +56,11 @@ interface CurrencyDenominationData {
     convertedValue: Derived<number>;
 }
 
+interface AttributeData {
```
zithith, line 59: This could probably be moved up to somewhere more universal. I know there are a couple of places in the roll calculations for items where I have in-lined this type definition

zithith (approved):

MangoFVTT (approved):

src/system/data/actor/common.ts
stanavdb: I could only find one other place this type was defined (see latest commit). Let me know if you have any I missed

zithith: I think that's it. I got a little mixed up between where I had basically defined the same type, but it's inferred as inline variables (see https://github.com/the-metalworks/cosmere-rpg/blob/075d82493e7c3b0ad3be4e47ebccfe7599bf5479/src/system/documents/item.ts#L1216 and the skill test function above it). So this probably won't help in those instances, but it's helpful to have available in case someone needs to refactor that. There's also something similar with the Skill object up on line 391 of that file too.

zithith (approved):

PR #221, opened 2025-02-08
feat(dice): implement dialog to select plot die roll result

### LINKED ISSUE #107: Improve handling of [dis]advantage mechanic
Advantage and Disadvantage are a little more intricate than keep highest or lowest (though in cases for the d20 and damage die it'd make do).
Currently plot die do not implement [dis]advantage mechanic as the logic is complex and the user needs to make some level of decision. And damage die do not increase the size of the pool (again this is due to it being a pool of dice inferring complexity to the situation). d20s currently use keep high/low functionality which is practical, but not entirely correct.
This lack of implementation is hampering roll automation.
https://discord.com/channels/1245805811087704085/1267574016311562361/1267587839252631595 has a good breakdown of the mechanic from a Brotherwise employee.

* Update the roller interface to allow advantage to be applied to damage die individually.
* Update the Roller to build up a number of rolls in place of a single pool.
* Once the roll is complete but before producing the chat message, provide a pop-up dialog to the roller or GM (depending on type of vantage) with the current results and a select which of each die pairing should be used.
* Feed these selections into the total calculation that exists.

While the possibility of complicated instances of 'vantage on damage dice are unlikely, trying to apply a simpler solution would mean if such edge cases did appear in a game they would be rolled wrong, not just uncatered-for. As we would have no logical alternatives for the plot die, we might as well just re-use this approach.

Display-wise, the selection could maybe be made within the chat message itself rather than a second dialog? I don't know whether that would be easier to implement or not, open to discussion. I also considered opening a pop-up for GMs ahead of the roll to define where they wanted to apply disadvantages, but it seemed like an unnecessary extra stage, when the players can discuss this around the table first and trust the person rolling will select the correct settings.

**Version**  
0.1.1

**Checklist:**
- [ ] I have searched the existing issues to ensure this feature has not already been requested.

### LINKED ISSUE #231: Implement advantage dice picking functionality for plot die
plot die can't have advantage applied to them at the moment as even the standard kh and kl modifiers don't adequately reflect how the advantage system works in CosRPG.

Apply a custom modifier to the plot die class so that when advantage is applied to it, an extra die can be rolled and the user is prompted to pick a result.
* Do so in a generic enough way that this could later be applied to any die in the pool.
* include differentiation between advantage and disadvantage, though for now we don't need to implement the prompt being sent to anyone but the rolling user.

stanavdb:
**Type**  
- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
This PR adds the ability to use advantage/disadvantage with plot dice by adding a new dice term "pick" modifier (`p`). When evaluated, this modifier brings up a dialog window to pick a number of dice to keep.  
Also adds "gm pick" (`gmp`) modifier, though resolves the same as the "pick" modifier for now.

**Related Issue**  
Closes #231 

**How Has This Been Tested?**  
1. Open an actor sheet with a weapon
2. Bring up the attack configuration dialog
3. Select "Raise the stakes" and apply advantage (or disadvantage) to the plot die
4. The dice result pick dialog comes up, pick a result
5. The chosen result is kept, the other discarded

image
image

**Checklist:**  

### DIFF
--- a/src/system/applications/dialogs/pick-dice-result.ts
+++ b/src/system/applications/dialogs/pick-dice-result.ts
@@ -110,6 +110,15 @@ export class PickDiceResultDialog extends ComponentHandlebarsApplicationMixin(
         // Get selected result
         const result = this.rolls[index];
 
+        // Ensure the amount picked is less than the amount to pick
+        if (this.picked.length >= this.data.amount && !!result.discarded) {
+            return void ui.notifications.error(
+                game.i18n!.format('DIALOG.PickDiceResult.Error.TooManyPicked', {
+                    max: this.data.amount,
+                }),
+            );
+        }
+
         // Toggle discarded
         result.discarded = !result.discarded;
 
--- a/src/system/applications/dialogs/pick-dice-result.ts
+++ b/src/system/applications/dialogs/pick-dice-result.ts
@@ -1,0 +1,194 @@
+import { AnyObject } from '@system/types/utils';
+
+// Dice
+import { PlotDie } from '@system/dice/plot-die';
+
+// Mixins
+import { ComponentHandlebarsApplicationMixin } from '@system/applications/component-system';
+
+const { ApplicationV2 } = foundry.applications.api;
+
+// Constants
+import { IMPORTED_RESOURCES } from '@system/constants';
+
+export namespace PickDiceResultDialog {
+    export interface Data {
+        /**
+         * The term for which to pick results
+         */
+        term: foundry.dice.terms.DiceTerm;
+
+        /**
+         * The amount of dice to pick
+         */
+        amount: number;
+    }
+}
+
+export class PickDiceResultDialog extends ComponentHandlebarsApplicationMixin(
+    ApplicationV2<AnyObject>,
+) {
+    /**
+     * NOTE: Unbound methods is the standard for defining actions and forms
+     * within ApplicationV2
+     */
+    /* eslint-disable @typescript-eslint/unbound-method */
+    static DEFAULT_OPTIONS = foundry.utils.mergeObject(
+        foundry.utils.deepClone(super.DEFAULT_OPTIONS),
+        {
+            window: {
+                minimizable: false,
+                resizable: true,
+                title: 'DIALOG.PickDiceResult.Title',
+            },
+            classes: ['dialog', 'pick-dice-result'],
+            tag: 'dialog',
+            position: {
+                width: 400,
+            },
+            actions: {
+                'select-result': this.onSelectResult,
+            },
+        },
+    );
+    /* eslint-enable @typescript-eslint/unbound-method */
+
+    static PARTS = foundry.utils.mergeObject(
+        foundry.utils.deepClone(super.PARTS),
+        {
+            form: {
+                template:
+                    'systems/cosmere-rpg/templates/roll/dialogs/pick-dice-result.hbs',
+            },
+        },
+    );
+
+    private rolls: foundry.dice.terms.DiceTerm.Result[];
+    private submitted = false;
+
+    private constructor(
+        private data: PickDiceResultDialog.Data,
+        private resolve: (
+            results: foundry.dice.terms.DiceTerm.Result[] | null,
+        ) => void,
+    ) {
+        super({});
+
+        // Mark all results as discarded to begin with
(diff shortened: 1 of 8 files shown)


stanavdb:
@MangoFVTT I've done a very basic styling pass on this, leaving for you

src/system/applications/dialogs/pick-dice-result.ts
```diff
@@ -0,0 +1,195 @@
...
+            ui.notifications.warn(
+                game.i18n!.format(
+                    'DIALOG.PickDiceResult.Warning.PickedLessThanTotal',
+                    {
+                        picked: this.picked.length,
+                        total: this.data.amount,
+                    },
+                ),
+            );
+        }
```
zithith, line 136: Why are you using a toast notification? Couldn't we just show a text counter in the window and disable the submission if not all are selected?

zithith (commented):
Love this! Great work!
One question commented in the file, and one change to the PR - I'd say this is only a partial closure of the issue. While this is clearly applicable approach to both d20s and damage too, it hasn't been applied to them yet

src/system/applications/dialogs/pick-dice-result.ts
stanavdb: We could, but I'm not a fan of locking things down all the way. With this set up if a user wants to select less than the available amount for some reason, they can (without having to adjust anything).

Maybe we get rid of the notification and change the text to "pick up to x"?

stanavdb:
> I'd say this is only a partial closure of the issue. While this is clearly applicable approach to both d20s and damage too, it hasn't been applied to them yet

Maybe this warrants splitting up the issue? While this approach would work for any other dice, I imagine a lot of people would simply like to use kh/kl, which would mean we'd have to make it configurable through settings. That's not critical for 1.0, so we shouldn't be including it in the 0.3 milestone.

zithith:
> > I'd say this is only a partial closure of the issue. While this is clearly applicable approach to both d20s and damage too, it hasn't been applied to them yet
> 
> Maybe this warrants splitting up the issue? While this approach would work for any other dice, I imagine a lot of people would simply like to use kh/kl, which would mean we'd have to make it configurable through settings. That's not critical for 1.0, so we shouldn't be including it in the 0.3 milestone.

we could split it, sure. In fact I might start looking at using Sub-Issues as that's a thing now.
However I think if we're only supporting one approach for 1.0 it should be to use the pick modifier by default across the board. Primarily it's the true way the rules are worded, but also because that will provide consistency between rolls.

MangoFVTT:
> > > I'd say this is only a partial closure of the issue. While this is clearly applicable approach to both d20s and damage too, it hasn't been applied to them yet
> > 
> > 
> > Maybe this warrants splitting up the issue? While this approach would work for any other dice, I imagine a lot of people would simply like to use kh/kl, which would mean we'd have to make it configurable through settings. That's not critical for 1.0, so we shouldn't be including it in the 0.3 milestone.
> 
> we could split it, sure. In fact I might start looking at using Sub-Issues as that's a thing now. However I think if we're only supporting one approach for 1.0 it should be to use the pick modifier by default across the board. Primarily it's the true way the rules are worded, but also because that will provide consistency between rolls.
> 
> Alternatively we make `p` the default and then keep the `kh`/`kl` approaches for those quick modifier key options that Mango did?

That would be my suggestion yeah:

- Click on dice -> pop up dialog to pick dice
- Shift-Click on dice -> kh, no dialog
- Ctrl-Click on dice -> kl, no dialog

stanavdb:
Should we then make it configurable through settings, so users can choose to default their d20/damage to kh/kl? Or maybe world level so the GM can configure it?

src/system/applications/dialogs/pick-dice-result.ts
zithith: Yeah, OK, that is fair. I can see there being cases of the players/GM realising they applied too many advantage or something so try to retcon it the simplest way by not picking all the options.

But the flipside of picking *more* than was calculated should be far less regular I think?

zithith:
I'll find time tomorrow or monday to split up the issue into some sub-issues so we can scope these PRs appropriately

zithith (approved):
Now scoped to #231 I'm happy to approve 😊

PR #240, opened 2025-02-12
Feat: Enrichers First Pass

### LINKED ISSUE #190: Custom actions when rolled do not display damage roll or give the opertunity for a advantage to be used on them
as reported by @Xonj on the metalworks discord.

**Describe the bug**  
A custom action item is made with basic action and skill test selected. the damage field is filled in. When going to roll the die portion of the info box is not present and you can not select advantage or disadvantage. 

**To Reproduce**  
1. Create a new custom action
2. set to basic action and skill test. add a damage die calculation
3. add it to actor sheet and roll it.
4. see the error of the missing field. 

**Expected behavior**  
Damage die field should be visible and you should be able to pick advantage or disadvantage.

**Screenshots**  
Correct Weapon style roll
image

Incorrect Missing dice and the damage line 
image

**Foundry VTT Version**  
12.331

0.2.1

**Browser**  
Firefox

**Checklist:**

### LINKED ISSUE #233: Create Prosemirror enrichers
**What needs changing?**  
Add basic enrichers for object data lookup, and potentially some roll options and condition reference.
Use the 5e system as an example.
Lookup enricher format: `[[lookup @field]]` | `[[lookup @field option:1]]{fallback text}}`

**Fields to Include:**
* Item
  * Name
  * Charges
* Linked Actor
  * Name
  * Stats & derived values
  * Type (humanoid/animal/etc)
* Token Name
* Stretch Goal - Target Name
* _Out of scope, post 1.0 - obfuscating names?_

This will mostly be useful for data entry in the premium modules that provide pre-built Foundry objects. But anything that might need these dynamic fields will benefit (e.g. homebrew).

We could not and just make do with hard coded values and the static descriptions from the book, but this doesn't make for a very "Value-added" product

zithith:
**Type**  
- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
Introduces a scaffold for registering custom prosemirror enrichers, including some suggestions of 5e ones we could adapt, then implements the first enricher. This is the data lookup for the item in question, it's linked actor or the user's [first] currently selected target.
The enrichers get their data context passed in from the enrichHTML call (via the relativeTo options field), this I have done my best to standardise to just pass in the originating item, though in theory, for other locations we could use practically any Document with a little it of tweaking.

**Related Issue**  
Closes #233
Closes #190 

**How Has This Been Tested?**  
* Imported an item from the compendium
* Updated description to include various lookup keys for numerous fields, including adding some to the chat description
* Saved the prosemirror editor and checked the output in the Item description window
* Dropped the item onto a character sheet
* Checked the output in the actions & equipment list, and by editing the embedded item
* Activated the item and checked chat output

 Note: this has been done in various states of token states, targeting states, actor data availability, etc.

Incoming...

**Checklist:**  

**Additional context**  
There are currently some limitations to what's possible:
* Various fields won't change until re-enriched (i.e. window frame is re-opened).
* We can't insert []'s into the command as the regex currently stand, so looking up items within arrays is not a thing (thusly for now I have limited targets to just the first target in the array)
* Without a fall back the default value is just the keyPath if the key cannot be found.

This is also unlikely an exhaustive list of what we could put in the data lookup, but it's a good start


src/system/utils/generic.ts
```diff
@@ -231,12 +231,12 @@ export interface TargetDescriptor {
...
-    const targets = new Map();
+    const targets = new Map<string, TargetDescriptor>();
     for (const token of game.user!.targets) {
         const { name, img, system, uuid } = (token.actor as CosmereActor) ?? {};
-        const phy = system.defenses.phy.value ?? 10;
-        const cog = system.defenses.cog.value ?? 10;
-        const spi = system.defenses.spi.value ?? 10;
+        const phy = system?.defenses.phy.value ?? 10;
+        const cog = system?.defenses.cog.value ?? 10;
+        const spi = system?.defenses.spi.value ?? 10;
```
zithith, line 239: This covers our back in case there is a token without an Actor available.
I encountered this with the B9 PCs who had data issues, so no sheets present in the list, but I think this accounts for tokens without linked actor data? might be wrong. Either way this allows it to fail gracefully and won't break the getEnricherData function

src/system/utils/enrichers.ts
```diff
@@ -0,0 +1,558 @@
...
+ *
+ * @example Formulas used for DCs will be resolved using data provided to the description (not the roller):
+ * ```[[/test skill=dec dc=12+@actor.attributes.int]]```
+ * becomes
+ * ```html
+ * <a class="roll-action" data-type="test" data-attribute="pre" data-dc="15">
+ *   <i class="fa-solid fa-dice-d20"></i> DC 15 Deception test
+ * </a>
+ * ```
+ */
```
stanavdb, line 326: The examples all show the use as `[[/test ...`, however the actual implementation requires `[[test ...`. Took me a while to figure out why I could only get lookup working.

src/system/applications/actor/base.ts
```diff
@@ -53,6 +54,7 @@ export class BaseActorSheet<
             actions: {
                 'toggle-mode': this.onToggleMode,
                 'edit-html-field': this.editHtmlField,
+                'trigger-enricher': enricherAction,
```
stanavdb, line 57: I wonder if it wouldn't be better to use some kind of global event handler for this. Any document we want to have enricher support needs to register this action now. This also doesn't really feel correct from a SoC viewpoint.

src/system/documents/actor.ts
```diff
@@ -1007,6 +1008,63 @@ export class CosmereActor<
...
+        };
+
+        const targets = getTargetDescriptors();
+
+        return {
+            actor,
+            target: targets.length > 0 ? targets[0] : undefined,
+        } as const satisfies EnricherData;
+    }
+
```
stanavdb, line 1067: We should probably bring enricher data in line with roll data

src/system/utils/enrichers.ts
zithith: Yeah, sorry, i got the functionality working and forgot to go back and update the comments.
My plan is to get an updated set of instructions/examples up in the wiki or at least on a Google doc somewhere (for now) and then link out to that from the jsdoc comments to keep the size down.

src/system/applications/actor/base.ts
zithith: So, I don't really like that approach, as we'd then have a listener listening out to every mouse click the user is making in the app and running them through the filter function to see if they are an enricher to be actioned. This is basically how the 5e ones worked. You can't declare a global listener on the embedded html from within the generating code (or at least, when I tried I couldn't get the timing right), it just has to be at some top level (`document` for 5e) and listening to generic events..
So this seems bad from an optimisation PoV to me.

Plus it makes more sense to me to see the individual areas that we want to support enrichers explicitly state that as it's clearer we are intending it.

src/system/documents/actor.ts
zithith: In what sense? Not sure I follow...

stanavdb: Right now we both the roll data and the enricher putting out data about an actor, but each in their own format. Some fields accessible in roll formulas don't exist in the enricher data and thus can't be looked up. That data should probably all be the same, otherwise it would get confusing for users.

src/system/applications/actor/base.ts
stanavdb: Using actions we're still using a generic listener, just the one foundry uses by default. Performance impact of a global click listener is pretty much none, especially compared to all the canvas stuff foundry is doing.

> Plus it makes more sense to me to see the individual areas that we want to support enrichers explicitly state that as it's clearer we are intending it.

I think we'd want to support enrichers everywhere the user could put text. That's at least going to be the expectation for most users. It wouldn't be clear why they could or couldn't use enrichers in certain documents (or chat). Can we support enrichers in chat this way? (if we can I'm happy to go with the current approach)

stanavdb:
> Might be worth having a separate issue for this so that we can figure out whether we want a document to use for supplying data values or to just code in an exception for journals

Yeah seems like the way to go

src/system/documents/actor.ts
zithith: I went this way because as I started it I was using the RollData, but it felt like a completely different use case to me. The vast majority of what I was trying to add would have needed to be added into RollData and likely wouldn't ever be needed in defining roll formulas.
So that says to me that they are different data types.
Granted I've never delved into any one system so deeply before, so not really used this type of thing in anger, but it seems way more logical to me for us to just have clear docs that we can point the user to that say "here's the data structure for the variables you can use to make your roll formulas with, these relate to the actor making the roll; and here's the data structure of the variables you can use to reference out from an enricher, which will relate to a number of sources filtered by the top level field". And when an enrichment can have a formula added (see tests and damage), they will support the RollData in the relevant spaces too.

Granted it might be worth trying to make the naming conventions more standardised, and maybe if we do that we could perhaps have some level of inheritance to the structures where they overlap?

zithith:
consider this back in draft until we get a 1.0 branch

zithith:
> > Might be worth having a separate issue for this so that we can figure out whether we want a document to use for supplying data values or to just code in an exception for journals
> 
> Yeah seems like the way to go

New Issue: #308

src/system/applications/actor/base.ts
zithith: @stanavdb in light of the upcoming events system, can we leave this be as it will probably need to flip over to that whole new thing as soon as it's ready anyway?

zithith: Will be staying as is.
After 1.0.0 release (or before if there's time after the dependent system is done) will re-visit how these are handled and hopefully change it so that the outputted HTML from the enrichers just invokes a hook and then the event system can take over.

src/system/documents/actor.ts
zithith: per discussions in 30/04 meeting I will swap the actor key to be the output of the actor.getRollData call and we'll revamp the central data format later on

Khunkurisu (approved):
Everything seems to work as expected now after the last couple changes. Documentation seems adequate for the moment.

PR #246, opened 2025-02-13
feat(actor sheet): add biography, appearance, and notes to actor data and sheets

### LINKED ISSUE #41: Add bio, appearance, notes/journal to character sheet
The character sheet is still missing a bio, appearance, and notes/journal section. 
The character data model will need to be updated with these fields.

NoahJGersh:
**Type**  

- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
This PR updates the common actor data model to support rich biography, appearance, and notes fields. It also adds those three fields to the character sheet, under a new Notes tab which is shared between characters and adversaries.

**Data Model Changes**
- `CommonActorData`
  - `system.biography`: `undefined` -> `string`
  - `system.appearance`: `undefined` -> `string`
  - `system.notes`: `undefined` -> `string`

*No data migration required.*

**To-Do List**
- [x] Update the actor data schema
- [x] Add a notes tab, with a notes field, to the adversary sheet
- [x] Add a biography tab to the character sheet, with the biography and appearance fields
- [x] Migrate the Purpose and Obstacle fields to the new character biography tab
- [x] Add notes to the character goals tab
- [x] Localize new strings in English
- [x] Update prose mirror instances for proper editing capability
- [x] Merge in new style changes
- [x] Update style to support new fields   

**Related Issue**  
Closes #41 

**How Has This Been Tested?**  
1. Create a character (or use an existing one)
2. Open the character sheet
3. Navigate to the Notes tab (between Goals and Effects)
4. Edit the various fields, and test the visibility toggle (Biography and Appearance should be expanded by default if there's anything present)
5. Close the sheet, and reopen it, to confirm data persistence
6. Repeat with an adversary

**Screenshots**  

Layout of the notes tab:
a screenshot of the notes tab, with a biography, appearance, and notes field, on a character sheet

On an adversary:
the notes tab, this time shown on an adversary sheet

And while editing:
the notes tab replaced with an active prose mirror editor, rather than the fields themselves

**Checklist:**  


src/style/sheets/actor/module.scss
```diff
@@ -390,6 +390,112 @@
...
+            overflow: hidden auto;
+            margin-right: 0;
+
+            .editor-content {
+                position: unset;
+                min-height: 300px;
+                padding: 0.25rem 0.5rem;
+            }
+        }
+    }
```
MangoFVTT, line 497: This seems to be a duplicate of the prose mirror styling from item scss right? If it's the same, I would coalesce the two into the parent file that applies to both item and actor (sheet.scss)? Avoids duplicate css like this

src/style/sheets/actor/module.scss
```diff
@@ -390,6 +390,112 @@
...
+                padding: 0 0.25rem;
+            }
+
+            &.expanded {
+                > .header::before {
+                    transform: none;
+                }
+            }
+        }
+    }
```
MangoFVTT, line 446: This also seems to be duplicate styling but with slightly different naming of classes. I would do the same thing, coalesce this css into the common parent file and make the html elements named the same (i.e. description-list or use some different generic name for both)

MangoFVTT (changes requested):
The templates and scripts seems fine, but I recommend moving duplicate css into files further up the hierarchy so that we keep the CSS files as short as possible. Any small differences should be extracted to override in the children files.

src/style/sheets/actor/module.scss
NoahJGersh: I've opted to use the generic for both, seeing as it's technically more accurate. There might be some merge conflicts with Khun's description expansion setting work, but it should be trivial to resolve.

MangoFVTT (approved):
LGTM

zithith (approved):

PR #250, opened 2025-02-16
fix(item sheets): fix ctrl-s functionality on item descriptions

### LINKED ISSUE #241: Ctrl+S in description fields fails
**Describe the bug**  
When editing the description of any item, using ctrl+s to save the item will result in unusual behavior. Specifically, it *will* save the changes, but it then also re-opens the prosemirror editor with the pre-save text. There is then no way to cancel out of editing without saving the old text, losing the previous changes.

**To Reproduce**  
1. Create a talent item (any works this is just an example)
2. edit the main description (any description field works)
3. make a change
4. hit ctrl+s instead of clicking save
5. observe odd behavior

**Expected behavior**  
The hot key should close the editor and save changes, as it did before the description menus were updated to include the different variations.

**Screenshots**  
N/A

**Foundry VTT Version**  
12.331

Cosmere system 0.2.2 and newer

**Browser**  
Local foundry install

**Additional context**  
N/A

**Checklist:**

NoahJGersh:
**Type**  

- [X] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This PR restores ctrl-s functionality to item sheet descriptions, immediately closing the editor when the user saves its content.

**Related Issue**  
Closes #241 

**How Has This Been Tested?**  
1. Create any item
2. Go to edit all three descriptions
3. Enter some content
4. Hit ctrl-s (or cmd-s if on a Mac)
5. Confirm that the editor view closes, and you are returned to the main description view
6. Edit and save again to confirm that the values update as expected

**Checklist:**  

### DIFF
--- a/src/system/applications/item/base.ts
+++ b/src/system/applications/item/base.ts
@@ -7,7 +7,6 @@ import { ComponentHandlebarsApplicationMixin } from '@system/applications/compon
 import { TabsApplicationMixin } from '@system/applications/mixins';
 import { getSystemSetting, SETTINGS } from '@src/system/settings';
 import { DescriptionItemData } from '@src/system/data/item/mixins/description';
-import HandlebarsApplicationMixin from '@league-of-foundry-developers/foundry-vtt-types/src/foundry/client-esm/applications/api/handlebars-application.mjs';
 
 const { ItemSheetV2 } = foundry.applications.sheets;
 
@@ -67,13 +66,18 @@ export class BaseItemSheet extends TabsApplicationMixin(
 
     /* --- Form --- */
 
-    protected static onFormEvent(
+    protected static async onFormEvent(
         this: BaseItemSheet,
         event: Event,
         form: HTMLFormElement,
         formData: FormDataExtended,
     ) {
         if (event instanceof SubmitEvent) return;
+
+        if ((event.target as HTMLElement).className.includes('prosemirror')) {
+            await this.saveDescription();
+        }
+
         if (!('name' in event.target!)) return;
 
         if (this.item.isPhysical() && 'system.price.unit' in formData.object) {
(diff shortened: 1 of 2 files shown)


src/system/applications/item/base.ts
```diff
@@ -67,13 +66,19 @@ export class BaseItemSheet extends TabsApplicationMixin(
...
+    protected static async onFormEvent(
         this: BaseItemSheet,
         event: Event,
         form: HTMLFormElement,
         formData: FormDataExtended,
     ) {
         if (event instanceof SubmitEvent) return;
+
+        // Handle prose mirror saving via hotkey
+        if ((event.target as HTMLElement).className.includes('prosemirror')) {
```
zithith, line 78: Won't this be triggering a save whenever an event bubbles out of the prosemirror editors? 
Granted they seem fairly well isolated, but could we not just narrow the conditional a bit more to look specifically for the key press events for the Ctrl+s chord? Just want to minimise the issue with band-aid side effects as mango mentioned in discord

NoahJGersh: Good question. As far as I can tell, we're only registering events that are triggering specifically from the editor `onChange`. The only events I can see foundry firing off in their `HTMLProseMirrorElement` class are `change`/`load` on save, the latter of which only being valid on toggled elements, and click handlers for the buttons themselves. When it comes to form events, only the `change` event is relevant. The one concern is that the same event is fired by the save button, but given that they're performing the same, idempotent operation, that's not a problem.

But granted, I can update the conditional to at least check that it's a `change` event. Unfortunately, we can't exactly get the key press that triggered the event- or at least I couldn't find a way to do so, given the way the Foundry prose mirror implementation locks down event handling. The event object also looks identical between keypress and button click activations, so it's tricky to filter that one out.

I'm open to any suggestions, we can probably discuss further in the Discord.

Khunkurisu: I've been using this branch locally for data entry and can confirm it is functioning as intended. I haven't encountered any bugs or problems using it.

Khunkurisu (approved):
Works, no issues spotted with the code, haven't encountered bugs with a few hours using this branch.

PR #258, opened 2025-02-17
fix(talent tree): fix issue causing node prerequisite dialog to close when switching to type talent

### LINKED ISSUE #247: Selecting 'Talent' in Requirement Popup Menu Fails
**Describe the bug**  
When setting up talent requirements in a Talent Tree item using the new setup, using the popup menu results in an error about `params.talents` being undefined in console, and closes the popup window instantly with no user feedback.

**To Reproduce**  
1. Create a Talent Tree item.
2. Add at least two Talents to it.
3. Click a Talent and hit the + on requirements to open the dialogue.
4. Select "Talents" from the select menu.

**Expected behavior**  
If the popup dialogue is going to have a Talents option, it should obviously still work- probably with the same functionality as pre rework of the trees that moved prereqs to them instead of on the talents. Since there is now the rightclick -> add connection functionality in place, it's also fine to just remove this option from the select menu.

**Screenshots**  
N/A

**Foundry VTT Version**  
12.331

System version 0.3.0

**Browser**  
Local Foundry install

**Additional context**  
N/A

**Checklist:**

stanavdb:
**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This PR fixes a bug that caused the edit node prerequisite dialog of the talent tree to error out when switching to type "talent".

**Related Issue**  
Closes #247 

**How Has This Been Tested?**  
1. Open talent tree
2. Click on node
3. Add new prerequisite
4. Set type to talent
5. Drag in talent

**Checklist:**  


Khunkurisu (commented):
I noticed when testing this that connections aren't made to talents added as prereqs this way. Is this intentional? And if so, doesn't that mean removing the talent option from the popup would be more ideal of a solution if it doesn't interact with the tree at all?

stanavdb:
> I noticed when testing this that connections aren't made to talents added as prereqs this way. Is this intentional? And if so, doesn't that mean removing the talent option from the popup would be more ideal of a solution if it doesn't interact with the tree at all?

That is intentional, yeah. Managing the node connections when dealing with the prerequisites through those components gets complex quickly, so I've simply avoided it for the time being. Unfortunately we can't just remove the option, as to support nested talent trees we have to be able to define prerequisite talents that aren't themselves part of the tree.

Khunkurisu:
> That is intentional, yeah. Managing the node connections when dealing with the prerequisites through those components gets complex quickly, so I've simply avoided it for the time being. Unfortunately we can't just remove the option, as to support nested talent trees we have to be able to define prerequisite talents that aren't themselves part of the tree.

Alright, thanks for clarifying

Khunkurisu (approved):
Functions as intended, no regressions or bugs found.

src/lang/en.json
```diff
@@ -584,7 +584,9 @@
                 },
                 "Node": {
                     "Prerequisites": {
-                        "Label": "Prerequisites"
+                        "Label": "Prerequisites",
+                        "OneOf": "One of",
+                        "CannotEditManaged": "Cannot edit managed prerequisite"
```
zithith, line 589: I'd be tempted to say "a system generated prerequisite" in the user message. Took me a while to realise that just reading the PR (I am right, this just means "the link we made automatically when you dragged the graphical connection"?).

Khunkurisu: When adding the talent requirement by clicking "add connection" this appears, yes.

stanavdb: > I'd be tempted to say "a system generated prerequisite" in the user message. Took me a while to realise that just reading the PR (I am right, this just means "the link we made automatically when you dragged the graphical connection"?).

Happy to change that for some added clarity

zithith (approved):

PR #265, opened 2025-02-19
fix(active effects): apply formula total to value rather than evaluated formula

### LINKED ISSUE #259: Active Effects Fail to Process Arithmetic
**Describe the bug**  
Active Effects return `NaN` when you set the `Effect Value` field to an arithmetic expression. This can be as simple as two number constants (ex `2 + 2`) or when including actor data (ex `2 + @attr.awa`).

**To Reproduce**  
1. Create an Active Effect on an item
2. Set the `Attribute Key` to any valid option (ex `system.resources.inv.max.bonus`)
3. Set the `Effect Value` to `2 + 2`
4. Add item to actor and toggle effect on (if necessary)

**Expected behavior**  
The key should be adjusted by 4, rather than being set to NaN.

**Screenshots**  

Image

Image

**Foundry VTT Version**  
12.331

|**Name**|**Version**|
|---|---|
|**System**|_Tested on 0.2.2 and 0.3.0_|

**Browser**  
Local install and Chrome

**Additional context**  
I am unsure if this is a system implementation bug or if Active Effects are not meant to allow arithmetic, but either way this is undesirable behavior. (It is however my understanding that AEs should support any arithmetic and JS Math functions that dice rolls do.)

**Checklist:**

stanavdb:
**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This PR fixes a bug causing math expressions in active effect values to result in `NaN`.

**Related Issue**  
Closes #259 

**How Has This Been Tested?**  
1. Open character sheet
2. Add active effect
3. Set attribute key to `system.resources.inv.max.bonus` and value to `1 + 1`
4. Investiture max increases by 2

**Checklist:**  

### DIFF
--- a/src/system/documents/active-effect.ts
+++ b/src/system/documents/active-effect.ts
@@ -21,7 +21,7 @@ function tryApplyRollData(actor: Actor, value: string): string {
         const data = actor.getRollData() as AnyObject;
 
         // Treat the change value as a formula and evaluate it
-        value = new Roll(value, data).evaluateSync().formula;
+        value = new Roll(value, data).evaluateSync().total.toString();
         return value;
     } catch {
         return value;


Khunkurisu (approved):
Works as expected and found no regressions.

PR #277, opened 2025-02-24
feat(chat): shift click damage buttons for modifier #276

### LINKED ISSUE #276: Shift-Click Damage Buttons for Modifier Popup
There are instances where damage was rolled incorrectly or should be adjusted in a way the system doesn't support, etc.

Add a popup dialogue when shift-clicking one of the damage application buttons (including healing and focus) that lets you input a positive/negative integer which will be added to the final applied value. (So after deflect and other modifiers occur.)

People can just manually increase/decrease the amount of health lost/gained after the fact, but this option would work with the chat cards nicely.

**Version**  
0.3.0

**Additional context**  
N/A

**Checklist:**
- [X] I have searched the existing issues to ensure this feature has not already been requested.

Khunkurisu:
**Type**  

- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
This adds a simple DialogV2 popup when shift-clicking one of the damage/healing/focus buttons of a damage roll chat card. For damage and healing, a positive or negative integer can be entered. Positive will increase damage/healing taken, negative will reduce. For focus, the integer must be 0 or greater. When used with one of the damage multipliers, the modifier is applied before the multiplier (so +4 to 8 damage will result in 24 when clicking the x2 button, as it will modify the input to 12 prior to the doubling).

**Related Issue**  
Closes #276.

**How Has This Been Tested?**  
1. Perform a test that includes a damage roll, such as by equipping and striking with a weapon.
2. Select a token
3. Hold shift and click one of the application buttons
4. Enter a valid number for the button type clicked
5. Hit enter/click confirm
6. Observe the appropriate modifications applying (in the damage card for damage/healing, and in the sheet for all 3 types)

image
image
image
image

**Checklist:**  

**Additional context**  
I made sure to use localization strings where applicable. If the placement of anything (code or localization stuff) is erroneous, just let me know what to fix and I'll sort it out. I did *not* implement any css class for the input box, instead doing inline styling. This can easily be rectified, but that was a whole can of worms I did not want to touch without supervision.

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -1212,6 +1212,10 @@
             "name": "Skip Roll Dialog by Default",
             "hint": "If enabled, rolls will skip the roll configuration dialog when clicked, using the default roll configuration. The dialog can still be accessed by holding the 'Skip/Show Dialog' key modifier from the system controls."
         },
+        "skipDamageModDialogByDefault": {
+            "name": "Skip Damage Modifier Dialog by Default",
+            "hint": "If enabled, the damage/healing card in chat will skip the modifier dialog when clicked, using the value shown in the card directly. The dialog can still be accessed by holding the 'Skip/Show Dialog' key modifier from the system controls."
+        },
         "enableOverlayButtons": {
             "name": "Enable Overlay Buttons",
             "hint": "When outputting a skill test or damage roll, enable overlay buttons on the chat message that offer additional functionality for the roll."
@@ -1248,7 +1252,6 @@
         "skipDialogDefault": "Skip/Show Dialog",
         "skipDialogAdvantage": "Skip Dialog with Advantage",
         "skipDialogDisadvantage": "Skip Dialog with Disadvantage",
-        "skipDialogRaiseStakes": "Skip Dialog with Raise Stakes",
-        "modifyDialogDamageCard": "Show Dialog to Modify Damage Card Value"
+        "skipDialogRaiseStakes": "Skip Dialog with Raise Stakes"
     }
 }
--- a/src/system/documents/chat-message.ts
+++ b/src/system/documents/chat-message.ts
@@ -840,9 +840,11 @@ export class CosmereChatMessage extends ChatMessage {
         event.stopPropagation();
 
         const button = event.currentTarget as HTMLElement;
-        const promptModify = areKeysPressed(
-            KEYBINDINGS.MODIFY_DIALOG_DAMAGE_CARD,
-        );
+        const promptModify =
+            !game.settings?.get(
+                'cosmere-rpg',
+                'skipDamageModDialogByDefault',
+            ) || areKeysPressed(KEYBINDINGS.SKIP_DIALOG_DEFAULT);
         const action = button.dataset.action;
         const multiplier = Number(button.dataset.multiplier);
 
--- a/src/system/settings.ts
+++ b/src/system/settings.ts
@@ -8,7 +8,8 @@ import { setTheme } from './utils/templates';
 export const SETTINGS = {
     INTERNAL_FIRST_CREATION: 'firstTimeWorldCreation',
     INTERNAL_LATEST_VERSION: 'latestVersion',
-    ROLL_SKIP_DIALOG_DEFAULT: 'skipRollDialogByDefault',
+    DIALOG_ROLL_SKIP_DEFAULT: 'skipRollDialogByDefault',
+    DIALOG_DAMAGE_MODIFIER_SKIP_DEFAULT: 'skipDamageModDialogByDefault',
     CHAT_ENABLE_OVERLAY_BUTTONS: 'enableOverlayButtons',
     CHAT_ENABLE_APPLY_BUTTONS: 'enableApplyButtons',
     CHAT_ALWAYS_SHOW_BUTTONS: 'alwaysShowApplyButtons',
@@ -65,12 +66,13 @@ export function registerSystemSettings() {
         });
     });
 
-    // ROLL SETTINGS
-    const rollOptions = [
-        { name: SETTINGS.ROLL_SKIP_DIALOG_DEFAULT, default: false },
+    // DIALOG SKIP SETTINGS
+    const dialogOptions = [
+        { name: SETTINGS.DIALOG_ROLL_SKIP_DEFAULT, default: false },
+        { name: SETTINGS.DIALOG_DAMAGE_MODIFIER_SKIP_DEFAULT, default: true },
     ];
 
-    rollOptions.forEach((option) => {
+    dialogOptions.forEach((option) => {
         game.settings!.register(SYSTEM_ID, option.name, {
             name: game.i18n!.localize(`SETTINGS.${option.name}.name`),
             hint: game.i18n!.localize(`SETTINGS.${option.name}.hint`),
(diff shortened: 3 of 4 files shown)


src/system/documents/chat-message.ts
```diff
@@ -837,18 +838,43 @@ export class CosmereChatMessage extends ChatMessage {
         event.stopPropagation();
 
         const button = event.currentTarget as HTMLElement;
+        const shiftHeld = event.shiftKey;
```
MangoFVTT, line 841: We shouldn't use directly the ShiftKey here imo. Use one of the keybindings that can be set. I would say whichever key is set to "Skip/Show Dialog" is the best option. Hardcoding shift/ctrl/alt modifiers in general should not be a thing when the framework for modifiable keybindings is already in place. You can refer to the code for clicking to roll an item with keybindings

src/system/documents/chat-message.ts
```diff
@@ -837,18 +838,43 @@ export class CosmereChatMessage extends ChatMessage {
...
+                            callback: (event, button, dialog) => {
+                                const element: HTMLInputElement =
+                                    button.form?.elements.namedItem(
+                                        'modifier',
+                                    ) as HTMLInputElement;
+                                return new Promise((resolve) =>
+                                    resolve(element.valueAsNumber),
+                                );
+                            },
+                        },
```
MangoFVTT, line 869: Judging by the screenshots, this does not create a dialog that is consistent with the other dialogs we have so far. For one thing, hardcoding styling into the element is bad™. For another, the width and layout are all off. 

Refer to the Short Rest dialog code in short-rest.ts to get an idea of how best to set this up so it is consistently styled. Either that, or move it to a template, which might actually be better since you are using this same content twice here and duplicate code is bad™.

src/system/documents/chat-message.ts
```diff
@@ -861,12 +887,36 @@ export class CosmereChatMessage extends ChatMessage {
...
+                    })) ?? 0;
+            }
             await Promise.all(
                 Array.from(targets).map(async (t) => {
                     const target = (t as Token).actor as CosmereActor;
                     return await target.update({
                         'system.resources.foc.value':
-                            target.system.resources.foc.value - 1,
+                            target.system.resources.foc.value -
+                            Math.max(1 + modifier, 1),
```
MangoFVTT, line 919: Unsure about the Math.max here. If someone decides to put in something like -2 as the modifier, wouldn't it be fine to simply allow it as a band aid means to increase focus through the chat buttons?

src/lang/en.json
```diff
@@ -1141,7 +1146,8 @@
             "Cancel": "Cancel",
             "Update": "Update",
             "RemoveFavorite": "Remove Favorite",
-            "Favorite":  "Favorite"
+            "Favorite":  "Favorite",
+            "Confirm": "Confirm"
```
MangoFVTT, line 1150: This should probably go under GENERIC.Button instead of here

MangoFVTT (changes requested):

src/system/documents/chat-message.ts
Khunkurisu: The button in the chat has -1 on it and otherwise only ever reduces focus, so I wanted to keep the modifications as-is. There’s a number of abilities that increase focus after use so a +focus button might be worth including or a generic focus modifier one. Dunno, I can remove the limit it just felt unintuitive for the button that says -1 to ever increase (and you’d need to put in -2 to do it).

src/system/documents/chat-message.ts
Khunkurisu: Yeah I agree with all of this, I shouldn’t have rushed and not done this with a template.

src/system/documents/chat-message.ts
Khunkurisu: This is smart. I disagree that hardcoding modifiers is generically bad, but I understand if the ability to change it is there then using it makes sense. Didn’t occur to me at the time, thanks!

MangoFVTT: It's not that it's generically bad, it's that within the same workflow we would have hardcoded and non-hardcoded, and people would get confused on whether or not their keybindings apply and where

src/system/documents/chat-message.ts
Khunkurisu: I've updated this with a template and such, styling is proper. I copied and modified the code from short rest.
image

src/system/documents/chat-message.ts
Khunkurisu: I went ahead and removed the minimum and check after talking more. If we decide later to adjust how the focus button works or add one for increasing, we can just refactor this to make sense with those changes.

src/system/documents/chat-message.ts
Khunkurisu: Added a new keybind that defaults to shift (double-checked that the overlap didn't cause problems, since they're in their own contexts it seems fine). Wanted people to be able to change it to something else if shift-clicking wasn't their desired means, without it also affecting the existing keybinds.

src/system/settings.ts
```diff
@@ -184,6 +185,10 @@ export function registerSystemKeybindings() {
             name: KEYBINDINGS.SKIP_DIALOG_RAISE_STAKES,
             editable: [{ key: 'KeyQ' }],
         },
+        {
+            name: KEYBINDINGS.MODIFY_DIALOG_DAMAGE_CARD,
+            editable: [{ key: 'ShiftLeft' }, { key: 'ShiftRight' }],
+        },
```
MangoFVTT, line 191: I'm of the opinion that it doesn't need to be a separate keybinding and can just use the generic Skip/Show Dialog. For a couple of reasons:

1. It keeps the behaviour of modifier keys relatively consistent across the application
2. It avoids us having to add a new keybinding setting every single time we do something with modifiers
3. If someone wants to change what their alt/shift/ctrl key does, they only need to change one setting and not several
4. If we separate keybindings too much we run the risk of key modifiers clashing with each other (e.g. if shift key shows dialog in one case but skips dialog in another, it may cause funky behaviour)

src/templates/chat/dialogs/damage-modifier.hbs
```diff
@@ -0,0 +1,8 @@
+<form>
+    <div class="form-group">
+        <label>{{ localize "GENERIC.Modifier"}}</label>
+        <div class="form-fields">
+            <input name="modifier" type="number" step="1" placeholder="+0" autofocus>
+        </div>
+    </div>
+</form>
```
MangoFVTT, line 8: This looks good off the top of my head, but make sure there isn't any funky styling due to a missing class or some such. Compare it to other dialogs would be my recommendation

src/system/settings.ts
Khunkurisu: I don't disagree with your points about binding creep necessarily, except that I absolutely do not think the two keybinds should be the same. They're two entirely different functions in different contexts, and it feels incredibly wrong to have them tied together.

src/templates/chat/dialogs/damage-modifier.hbs
Khunkurisu: Including button hover and cursor focus
image

src/system/settings.ts
Khunkurisu: The only way I can personally see the skip/show dialog keybind being appropriate for this context is if the dialog appearing is the default behavior, and much like the roll dialog, we have a setting to allow it to be skipped by default.

Khunkurisu: Or the inverse. Have it skipped by default but a setting to show it by default. Then the keybind being "skip/show dialog" works as a catch-all.

Khunkurisu:
@MangoFVTT still needs rereview of your requested changes when you can manage it!

MangoFVTT (approved):

PR #290, opened 2025-03-24
feat: roll config styling and damage dice advantage mechanics

### LINKED ISSUE #232: Apply advantage mechanic to all die and allow settings for users to set defaults
Currently our implementation of advantage isn't exactly right. The mechanic is more nuanced than the kh/kl modifiers. We need this to be handled correctly as the default, but allow user flexibility.

* Take the work from #231 and apply the custom modifier(s) to any die type
* Add settings for the user to select whether they default to kh/kl or picking with non-plot die, with the system default being to pick
* Update the roller dialog so that it can handle multiple [dis]advantages (only has to be one or the other) and spread them across d20, plot and damage die
  * Design suggestion before was to display the pool as a set of die icons.
  * Then you could click the icons to apply one of the advantages to that die.
  * [Zithith's Suggestion]: Still split the pool up by roll term - this will identify plot d6s from standard d6s in damage easier.
    * This then gives an alternative input for advantages by way of a number or slider input for each line (which may be simpler to manipulate in code).
  *  There will need to be a max number displayed of advantages to apply. Possibly a counter: "X/Y applied".
* Add the relevant number of die to each part of the pool and use the custom modifier to prompt for the right number of picks per die term (or use kh/kl per the setting selection).
  * Note: whether it's the player or GM who picks for adv/disadv will change depending on who is rolling!
  * Though we still will just implement this as the player picking regardless of whether it's adv or disadv.
* Leave the shortcut modifier keys as they are (they can be shortcuts to just rolling the skill check as if it were a D&D check)

For 1.0 we will have to put up with multiple prompts appearing if the advantage is applied to more than one type of die term. If we can come back to it later it'd be great to try and amalgamate all the picks needed for a roll but I don't know if that's possible with roll modifiers.

If using the "click a die icon to apply advantage" approach then modifier keys could then be used here again to define if the user wanted it as "pick" or kh/kl. Implementing that isn't the scope of this issue though.

MangoFVTT:
**Type**  

- [ ] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  
A PR that refactors and restyles the roll configuration dialog, allowing for advantage/disadvantage to be applied to individual dice.  Also includes refactors to roll config data so that damage dice can have advantage/disadvantage applied to individual dice, and fixes a number of issues that crop up as a result of this new functionality (e.g. retroactive crits). 

**Related Issue**  
Closes #232.

**How Has This Been Tested?**  
Tested various roll configs and ensured roll flow continues to work.

image

**Checklist:**  


src/system/applications/item/dialogs/item-consume.ts
stanavdb, line 1: Not entirely familiar with the DialogV2 lifecycle, doesn't this need a close handler that resolves the promise to cancel the consumption?

MangoFVTT: @stanavdb As far as I am aware no, the default behaviour is resolving with null which elegantly ends the flow without any consumption. I think you only need a close handler if you want it to do something specific on close in DialogV2

Just to be sure, I did some tests just now with closing the window and there are no errors, no consumption, and no issues that I can see with the behaviour

stanavdb: We create the promise ourselves though, so if you close at the moment it's simply never resolved.

MangoFVTT: There isn't any unresolved promise right? The dialog works on callbacks and the default callback of closing the dialog returns undefined. The outcomes are:
- Player presses continue, promise resolves to shouldConsume = true/false depending on selected checkbox, consume goes through and item roll continues
- Player presses close, promise resolves to undefined result, which gracefully ends the roll flow

MangoFVTT: I can add a manual _onClose but I don't think it actually makes any difference since it would do the same thing as the parent it inherits, i.e. return resolve(null)

stanavdb: Is it resolving when closing? If you can test that might be the best way to confirm

MangoFVTT: You're right, I was under the impression that the _onClose inherited function gracefully resolved the dialog, but it does not. I've pushed a fix. I also fixed it for the ShortRest dialog which I noted had the same issue

stanavdb (approved):
Looks good

PR #292, opened 2025-03-25
feat(conditions): add basic cumulative condition support

### LINKED ISSUE #48: Add support for cumulative conditions
Some conditions are able to be stacked and have cumulative effects. The only 2 I see are Enhanced and Exhausted, but there may be more in the future. Currently conditions are a binary on/off and can't represent the higher stages.

stanavdb:
**Type**  
- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
This PR adds basic support for cumulative conditions. Currently only set up to use with conditions and not custom Active Effects, though should be easy to adapt for general use. 

**Related Issue**  
Closes #48 

**How Has This Been Tested?**  
1. Open character sheet, navigate to effects tab
2. Click on exhausted condition
3. Condition should change to "Exhausted [-1]"
4. Click on exhausted condition again
5. Conditions should change to "Exhausted [-2]"
6. Right click exhausted condition twice
7. Condition should become inactive.

image

**Checklist:**  

**Additional context**  
This functionality really needs improved UX on the conditions tab, however I think it'd be good to include in 0.3.0 as as and improve UX down the line.


src/system/types/cosmere.ts
```diff
@@ -24,7 +24,6 @@ export const enum Condition {
     Determined = 'determined',
     Disoriented = 'disoriented',
     Empowered = 'empowered',
-    Enhanced = 'enhanced',
```
MangoFVTT, line 27: I assume we are removing Enhanced because it is handled by Item effects instead? I think we should also remove Encumbered since that is not a condition in the Handbook.

With that being said, by removing 2 conditions the grid will now be unbalanced with the last row only having 1 entry. Unless the data entry team has found new conditions we should add in as replacement, we'll need to consider a change to the UI here so that it doesn't look ugly as sin. Perhaps Exhaustion should be it's own thing in the sheet instead of under Conditions? or have a separate row under conditions. Not sure

src/system/documents/active-effect.ts
```diff
@@ -1,15 +1,123 @@
...
+                'system.count',
+            ) as number;
+            const config =
+                CONFIG.COSMERE.conditions[this.statuses.first() as Condition];
+
+            data.name = `${game.i18n!.localize(config.label)} [${
+                config.countDisplayTansform
+                    ? config.countDisplayTansform(count)
+                    : count
+            }]`;
```
MangoFVTT, line 88: I think putting the cumulative version in the name is fine as a temp solution, but as far as UI goes it's not particularly good. Instead, we should have Exhausted show a changing number next to it instead of the toggle (and a faded out 0 when it is inactive). Something to do for 1.0 though I think

src/system/documents/active-effect.ts
```diff
@@ -1,15 +1,123 @@
...
+import { AnyMutableObject, AnyObject } from '@system/types/utils';
+
+export class CosmereActiveEffect extends ActiveEffect<ActiveEffectDataModel> {
+    /* --- Accessors --- */
+
+    /**
+     * The number of stacked instances of this effect. Used for cumulative effects.
+     * Shorthand for `system.count`.
+     */
+    public get count() {
```
MangoFVTT, line 13: I think this might be clearer if it was called "stacks". While reading the review initially I wasn't immediately sure what count was referring to in regards to the condition

MangoFVTT (approved):
This looks good. I have a couple comments about the UI that should probably be a separate thing for 1.0

src/system/types/cosmere.ts
stanavdb: The reason I've dropped enhanched is that you can't just have the enhanched condition on its own. Enhanched always comes with an attribute and value (so Enhanced [+2 Speed] for example). So it doesn't make sense to have it as a condition.

Having another read through, we should drop Afflicted for the same reason (it always needs a die and damage type). Which means we can drop an even 3 conditions and keep the UI the same.

For 1.0 however we should have a think about how we can easily apply those type of conditions to an actor without needing to have an item set up with them before hand. Requiring items to have pre-setup active effects for base conditions is limiting on GMs.

src/system/documents/active-effect.ts
stanavdb: Yeah agreed. This is very a much a quick and dirty solution.

src/system/documents/active-effect.ts
stanavdb: I'm happy to change it to "stacks"

src/system/types/cosmere.ts
MangoFVTT: Fair point, i think we could work out something with having either a new section of effects or small buttons/some kind of interactivity near the individual attributes to allow for quickly adding afflicted/enhanced to a specific attribute.

zithith (approved):

MangoFVTT (dismissed):

MangoFVTT (approved):

PR #304, opened 2025-03-31
fix(0.2-0.3 migration): fix issue causing talent tree migrations to fail and improve migration robustness

### LINKED ISSUE #302: Failing migrations
**Describe the bug**  
Migrations are failing for some worlds. It appears to be related to custom talent trees though this is not confirmed. 

**To Reproduce**  
1. Add the following world: 
[oldaneg-crpg - Copy.zip](https://github.com/user-attachments/files/19524656/oldaneg-crpg.-.Copy.zip)
2. Open foundry
3. Load the world with 0.3.0

**Expected behavior**  
Actors and talent trees should be migrated over to the new version but they are not.

**Foundry VTT Version**  
12.331

|**Name**|**Version**|
|---|---|
|**System**|0.3.0|

**Checklist:**

stanavdb:
**Type**  
- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This PR solves the issue causing 0.2-0.3 migrations to fail if custom talent trees are present in the world. The previous implementation failed since in the type we were using a collection, however when we grab the raw data from Foundry we get back a plain object. This PR also generally makes the migration more robust by wrapping each individual document migration in its own try/catch.

**Related Issue**  
Closes #302 

**How Has This Been Tested?**  
1. Install the test world and backups shared in test data
2. Revert to the oldest back up
3. Open the world with 0.3
4. Migration now succeeds

**Checklist:**  

### DIFF
--- a/src/system/utils/migration/migrations/0.2-0.3.ts
+++ b/src/system/utils/migration/migrations/0.2-0.3.ts
@@ -13,6 +13,7 @@ import {
     getPossiblyInvalidDocument,
     getRawDocumentSources,
 } from '@system/utils/data';
+import { handleDocumentMigrationError } from '../utils';
 
 // Types
 import { Migration } from '@system/types/migration';
@@ -22,7 +23,11 @@ import { CharacterActorDataModel } from '@src/system/data/actor/character';
 // Constants
 import { SYSTEM_ID } from '@system/constants';
 import COSMERE from '@src/system/config';
-import { AnyMutableObject, AnyObject } from '@src/system/types/utils';
+import {
+    AnyMutableObject,
+    RawDocumentData,
+    AnyObject,
+} from '@src/system/types/utils';
 
 interface LegacyTalentTreeDataModel {
     nodes: Record<string, LegacyNode>;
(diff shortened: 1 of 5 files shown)


src/system/utils/migration/migrations/0.2-0.3.ts
```diff
@@ -169,53 +178,63 @@ async function migrateTalentTrees(items: AnyObject[]) {
...
+
+                if (actor.type === ActorType.Character) {
+                    /* --- Advancement ---*/
+                    if (
+                        isNaN((actor.system as CharacterActorDataModel).level)
+                    ) {
+                        foundry.utils.mergeObject(changes, {
+                            ['system.level']: 1,
+                        });
+                    }
```
NoahJGersh, line 217: Did we want to update the condition here to be a little bit more lenient? I've confirmed that everything migrates correctly on the test world, so I don't know that it's necessary, but it might be a good idea to catch some other falsy values.

NoahJGersh (approved):
Confirmed that everything works as intended on the test world, just one thought from the discord conversation surrounding the migration issues that were popping up.

src/system/utils/migration/migrations/0.2-0.3.ts
stanavdb: I didn't put it in, since I wasn't sure if we wanted to split that into its own PR or not. But I'm happy to include it here

NoahJGersh (approved):

PR #307, opened 2025-04-01
fix(component system): fix issue causing form associated components to lose track of their value

### LINKED ISSUE #306: Form associated components are not passing their value when another element is updated
**Describe the bug**  
The system uses custom from associated components so that they can tie into the existing form event flow. These components' value gets correctly passed to the form data when they are updated. However when any other form element (custom or default) gets updated, the value of these custom events becomes undefined. Causing these fields to be reset.

**To Reproduce**  
1. Create path
2. Add linked skills or talent tree
3. Update the identifier
4. Close and open item
5. Linked skills/talent tree are now empty.

**Expected behavior**  
The custom form associated components should correctly pass their state.

**Foundry VTT Version**  
12.331

|**Name**|**Version**|
|---|---|
|**System**|0.3.0|

**Checklist:**

stanavdb:
**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This PR fixes an issue causing our form associated components to "forget" their value when the form is submitted as result of anything but a change event from that component. 

**Related Issue**  
Closes #306 

**How Has This Been Tested?**  
1. Create path
2. Ad linked skills
3. Update the identifier
4. Close and open item
5. Linked skills remain as they were

**Checklist:**  

### DIFF
--- a/src/system/applications/component-system/system.ts
+++ b/src/system/applications/component-system/system.ts
@@ -150,8 +150,26 @@ export function registerComponent(
             class extends HTMLElement {
                 static formAssociated = true;
 
-                public value: unknown;
                 public name: string | undefined;
+                private _value: unknown;
+                private internals: ElementInternals;
+
+                constructor() {
+                    super();
+
+                    this.internals = this.attachInternals();
+                }
+
+                public set value(value: unknown) {
+                    this._value = value;
+                    this.internals.setFormValue(
+                        value as string | File | FormData | null,
+                    );
+                }
+
+                public get value() {
+                    return this._value;
+                }
             },
         );
     }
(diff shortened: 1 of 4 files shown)


zithith (approved):

Khunkurisu:
This doesn't seem to stop the skills from resetting when you add a talent tree. Nor the talent tree resetting when you adjust the skills.

stanavdb:
> This doesn't seem to stop the skills from resetting when you add a talent tree. Nor the talent tree resetting when you adjust the skills.

Looks like this was due to the values not getting assigned properly on initial render. If you opened the sheet and assigned both while open, any further assignments don't cause resetting. This is fixed with the last commit.

Khunkurisu:
@zithith you may want to rereview this since you did the previous one, but I can confirm that at least from testing while working with paths and talent trees this functions as expected.

zithith (approved):

PR #326, opened 2025-04-24
Feat(ux): skill rank alternative setting method #137

### LINKED ISSUE #137: Clearer UX/UI for setting skill bonus
Currently to set your skill bonus in a skill you need to left click to increase and right click to decrease the number of points. There is a tooltip explaining this but it might be a little confusing

Change the way skills are set to allow users to click on the relevant bullet to set their skill level at that rank (i.e. if they click the 4th bullet on a skill in sets that skill to be +4)

Alternatively could maybe try to implement something to make it clearer to the user what they have to do to set their skills

**Additional context**  
Suggested by `@deathpote_` on 21/10/2024

**Checklist:**

Khunkurisu:
**Type**  
- [ ] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  
This implements click-to-set functionality on skill rank pips. It adds a toggle in settings (off by default) to allow clients to use the legacy skill input method. It also provides a new tooltip for the functionality and fixes a localization issue with the existing tooltip. Edit: Now also implements a class for visualizing pips as beyond max rank for the actor.

**Related Issue**  
Closes #137 

**How Has This Been Tested?**  
1. Open actor sheet, click a skill pip, observe rank being set.
2. Right-click a skill pip, observe rank being reset to 0.
3. Click a skill pip that's already equal to the current rank and observe the skill rank decrement.
4. Toggle the legacy setting and observe that functionality return.

image
image

**Checklist:**  

**Additional context**  
I used a somewhat hacky method to get the index of the pips. If anyone knows a better way I'm happy to revert the helper I used and implement it in the suggested manner.


src/lang/en.json
```diff
@@ -1287,6 +1288,10 @@
 			"name": "Expand Item Description By Default",
 			"hint": "Determines whether the description section of an item should be expanded by default when opening the item sheet."
 		},
+		"skillIncrementDecrementToggle": {
+			"name": "(Legacy) Skill Pip Increment/Decrement Toggle",
```
MangoFVTT, line 1292: I don't think we need (Legacy) here, we can just leave the functionality as a regular setting. That said, usually I like to frame settings as a true/false question for clarity. So this should be Enable Skill Pip Increment Behaviour for example, so that the checkbox is a clear answer to the setting proposition.

src/system/applications/actor/components/skill.ts
```diff
@@ -67,15 +67,46 @@ export class ActorSkillComponent extends HandlebarsApplicationComponent<
     ) {
         event.preventDefault();
 
+        // Check if click is left or right mouse button
         const incrementBool: boolean = event.type === 'click' ? true : false;
+        // Check if the legacy behavior is toggled on
+        const shouldIncDec: boolean = game.settings?.get(
+            'cosmere-rpg',
+            'skillIncrementDecrementToggle',
+        ) as boolean;
```
MangoFVTT, line 76: Use getSystemSetting from the SettingsUtility instead of this, so that we don't have hardcoded IDs anywhere.

src/system/applications/actor/components/skill.ts
```diff
@@ -116,6 +147,10 @@ export class ActorSkillComponent extends HandlebarsApplicationComponent<
 
             editable: !this.readonly,
             pips: this.pips,
+            legacyMode: game.settings?.get(
+                'cosmere-rpg',
+                'skillIncrementDecrementToggle',
```
MangoFVTT, line 152: Same here

MangoFVTT (changes requested):
Looks good, just some minor things

src/system/applications/actor/components/skill.ts
MangoFVTT: `should be getSystemSetting(SETTINGS.<whatever you named it>)`

Khunkurisu:
Fixed the above issues and also actually provided the same functionality to adversary sheets because I forgot to previously.

src/system/applications/actor/components/adversary/skills-group.ts
```diff
@@ -57,19 +58,47 @@ export class AdversarySkillsGroupComponent extends HandlebarsApplicationComponen
...
+                false,
+            );
+        } else {
+            // Increment/Decrement the skill rank based on click type
+            await this.application.actor.modifySkillRank(
+                skillId,
+                incrementBool,
+                false,
+            );
+        }
```
MangoFVTT, line 101: I'm not intimately familiar with this side of the codebase but is there no way we can avoid the duplicate code here? It seems to me that the skill.ts script is common to both adversary and character so it should be able to have the code there and be used by both no?

MangoFVTT (changes requested):

src/system/applications/actor/components/adversary/skills-group.ts
Khunkurisu: It was in both previously so I assumed it would be necessary with my changes too.

Khunkurisu: I’ll test later if it can be removed or at least reduced to a single shared function.

src/system/applications/actor/components/adversary/skills-group.ts
```diff
@@ -57,19 +58,47 @@ export class AdversarySkillsGroupComponent extends HandlebarsApplicationComponen
...
-        );
+        if (!shouldIncDec) {
+            // Get the index of the clicked pip
+            const rankIndex: number = $(event.currentTarget!).data(
+                'index',
+            ) as number;
+            // Get current skill rank
+            const currentRank: number =
+                this.application.actor.system.skills[skillId].rank;
+            // Determine if rank and pip clicked match
```
zithith, line 81: I don't think we want to do this. If we aren't in the increment/decrement mode, clicking the same pip should do nothing imo.

src/system/applications/actor/components/adversary/skills-group.ts
zithith: I think there definitely should be a way to avoid duplicating. I'm just concerned that the adversary skill group is handling this event at all? might be a case of tweaking/refactoring the event handlers...

src/templates/actors/components/skill.hbs
```diff
@@ -12,30 +12,30 @@
...
+            {{#each (numberRange skill.rank)}}
                 <li class="pip active">
-                    <div></div>
+                    <div {{#if ../editable}}data-action="adjust-skill-rank" data-index='{{@index}}'{{/if}}></div>
                 </li>
-            {{/times}}
-            {{#times (sub 5 skill.rank)}}
+            {{/each}}
+            {{#each (numberRange (sub 5 skill.rank))}}
                 <li class="pip">
```
zithith, line 30: can you quickly add a css class here for indexes beyond the character's max skill rank?
Now we have the index here, it should just need passing in the max rank from the actor in the prepareContext and then a little `class="{{#if `...`{{/if}}"`

Khunkurisu: This is now done!

src/system/applications/actor/components/adversary/skills-group.ts
Khunkurisu: Removed this functionality in favor of keeping it simple: click = set to rank; rightclick = clear ranks

src/system/applications/actor/components/adversary/skills-group.ts
Khunkurisu: It turns out the duplicated code was entirely redundant (even before my changes). I simply removed the action and related function, and it works fine.

MangoFVTT (dismissed):
LGTM

zithith (approved):

MangoFVTT (approved):

Issue #320, opened 2025-04-23
Rename Darkvision


zithith:
**What needs changing?**  
As per johnny's comments on a set of weekly questions.

Khunkurisu:
Rename to "Sense"

PR #329, opened 2025-04-27
fix: replace placeholder combat tracker styling

### LINKED ISSUE #30: Replace placeholder icons in combat tracker
Combat tracker currently default foundry icons for the turnspeed and activation buttons. They are fine, but we should be able to find better options.

TheFlyingP1g has told us that Brotherwise will be making some Plotweaver specific resources available to us in the coming weeks. We should see if that includes any icons that would make for a good activation icon.

MangoFVTT:
**Type**  

- [ ] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  

- Replaces the placeholder combat tracker icons with new ones.
- Improves combat tracker styling a bit.
- Re-adds some conditions to the sheet so that they can be tracked as statuses. 
- Add some generic statuses to the token status window.

**Related Issue**  
Closes #30.

**Checklist:**  


stanavdb (commented):
No comments on the combat tracker stuff, but I feel the `sheet` flag for conditions is not the way to go. We should reserve Conditions for status effects specific to the rpg and keep a separate list of additional status effects. Then use both in `registerStatusEffects` during system initialisation.

MangoFVTT:
> No comments on the combat tracker stuff, but I feel the `sheet` flag for conditions is not the way to go. We should reserve Conditions for status effects specific to the rpg and keep a separate list of additional status effects. Then use both in `registerStatusEffects` during system initialisation.

I did consider that, but I decided on this method because it fluidly supports swapping conditions on/off from the sheet for both homebrew purposes and extensibility reasons for us. If someone wants, for example, to play a homebrew game that has the non-sheet statuses show up in the sheet because they've introduced rules for them, they can simply pass a CONFIG.COSMERE.conditions.X.sheet = true in a world script and everything will simply work without issue. It seemed a far more graceful solution than having an entirely separate list to be aware of and modify.

stanavdb:
> I did consider that, but I decided on this method because it fluidly supports swapping conditions on/off from the sheet for both homebrew purposes and extensibility reasons for us. If someone wants, for example, to play a homebrew game that has the non-sheet statuses show up in the sheet because they've introduced rules for them, they can simply pass a CONFIG.COSMERE.conditions.X.sheet = true in a world script and everything will simply work without issue. It seemed a far more graceful solution than having an entirely separate list to be aware of and modify.

I see where you're coming from however `sheet = true` isn't really a good delimiter between statuses and conditions. I'd like to keep a clear separation between what is a condition (a status effect that has rules associated with it) and a regular status (just a flag) so either way or module devs could use that in the future. 
Maybe we need to refactor `CONFIG.COSMERE.conditions` to something like `CONFIG.COSMERE.statuses` and define some statuses as conditions? We can still keep the `sheet` flag in that case, I just don't like _it_ being what separates a condition from a status.

src/system/config.ts
```diff
@@ -97,55 +97,107 @@ const COSMERE: CosmereRPGConfig = {
     },
 
     conditions: {
+        [Condition.Afflicted]: {
+            label: 'COSMERE.Conditions.Afflicted',
+            icon: 'systems/cosmere-rpg/assets/icons/svg/conditions/afflicted.svg',
+            sheet: true,
+        },
+        [Condition.Blind]: {
```
zithith, line 105: Let's call this "senses obscured".
Or we add a "deafened"? potentially "muted" too?
(technically if we go this route you'd want something for touch, taste and smell too, but I suspect they will be the less used options for primary senses and won't really have much effect outside of that)

zithith:
So, the changes look good to me barring one little thing.
I don't know where I stand on the conditions bit. My gut says the simplest and clearest solution is stans' last suggestion (which is basically the current system, just renamed for semantics).
But I wonder if we aren't approaching it from the wrong direction. Conditions are the things set out in the book. That's all they need to be really. Then we have "Character state we think that would be good to track on a token for at-a-glance reminders" and "Character state we think would be good to track on the sheet, for wider ranging effects". The latter of which really includes all the conditions from the book(s). The list needs to be extensible (which we get by just defining them as config arrays and providing a register call). To be fair, we have more like 3 states: Actual Conditions, State implied in the book with some rules (flying, hidden), and general RP things (burrowing, invisible).
Would it be clearer to keep those lists separate, or one big list that is typed? and is it just the type that limits whether it's on the sheet or should it be a specific toggle instead? Do we want a similar toggle for things to show on the token too? (side note, I've seen that it's possible to do token states that are driven by items added to the sheet but aren't selectable in the token context menu. Not sure it's applicable to Cosmere, but good to think about)

MangoFVTT:
> So, the changes look good to me barring one little thing. I don't know where I stand on the conditions bit. My gut says the simplest and clearest solution is stans' last suggestion (which is basically the current system, just renamed for semantics). But I wonder if we aren't approaching it from the wrong direction. Conditions are the things set out in the book. That's all they need to be really. Then we have "Character state we think that would be good to track on a token for at-a-glance reminders" and "Character state we think would be good to track on the sheet, for wider ranging effects". The latter of which really includes all the conditions from the book(s). The list needs to be extensible (which we get by just defining them as config arrays and providing a register call). To be fair, we have more like 3 states: Actual Conditions, State implied in the book with some rules (flying, hidden), and general RP things (burrowing, invisible). Would it be clearer to keep those lists separate, or one big list that is typed? and is it just the type that limits whether it's on the sheet or should it be a specific toggle instead? Do we want a similar toggle for things to show on the token too? (side note, I've seen that it's possible to do token states that are driven by items added to the sheet but aren't selectable in the token context menu. Not sure it's applicable to Cosmere, but good to think about)

I'm of the opinion it is better to keep them as one list because, while they are SEMANTICALLY different, they are FUNCTIONALLY the same exact things. They're all statuses that can apply to an actor, that's it. To me it makes no sense to arbitrarily split them into lists, force people to figure out which list to add them to, and have to do separate list iterations and all that janky stuff. 

I agree 100% with renaming things as per stan's suggestion tho

src/system/config.ts
Khunkurisu: Blind staying Blind makes sense to me cause it has direct foundry functionality (affects token vision). +1 for Deafened and Silenced (preferred name over muted).

Khunkurisu:
> > I did consider that, but I decided on this method because it fluidly supports swapping conditions on/off from the sheet for both homebrew purposes and extensibility reasons for us. If someone wants, for example, to play a homebrew game that has the non-sheet statuses show up in the sheet because they've introduced rules for them, they can simply pass a CONFIG.COSMERE.conditions.X.sheet = true in a world script and everything will simply work without issue. It seemed a far more graceful solution than having an entirely separate list to be aware of and modify.
> 
> I see where you're coming from however `sheet = true` isn't really a good delimiter between statuses and conditions. I'd like to keep a clear separation between what is a condition (a status effect that has rules associated with it) and a regular status (just a flag) so either way or module devs could use that in the future. Maybe we need to refactor `CONFIG.COSMERE.conditions` to something like `CONFIG.COSMERE.statuses` and define some statuses as conditions? We can still keep the `sheet` flag in that case, I just don't like _it_ being what separates a condition from a status.

This, to me, feels like one of the better options for making it clear to third party modules as well what has an actual *effect* on the character and what functions as a marker. It also then still provides the ability to add/remove from the sheet as needed.

src/system/config.ts
zithith: We should probably change that then, as this system doesn't function that way. At least, not semantically.
Johnny's already asked us to swap the name of Darkvision and use that to represent obscured senses (there's an issue on 3.1 for it) so if there is the same functionality we need to have it turn that sense on rather than just call it blind

zithith:
Cool, sounds like we have a plan then!

src/system/config.ts
MangoFVTT: Yeah it has to stay Blind because otherwise the automated vision removal does not work unfortunately. I had it as "obscured senses" originally but that breaks it

zithith: But my point is, in Cosmere, being blinded doesn't remove vision, so we don't want it doing that. If anything, we need it to switch the token vision into the Obscured Senses mode (read Darkvision), but I don't know if that's possible?

Khunkurisu: Being blinded does do that here though?? It reduces your vision radius to your Sense range. Which is something we can override the foundry “blind” status to do.

zithith: If that is something we can do, then yes, we can have a set of 3 statuses that should trigger the Foundry blinded effect and then in #320 set the blinded effect to switch the token to using "Obscured Senses" vision mode

MangoFVTT: It should be noted that "Blinded" in foundry does NOT revert you to darkvision, it outright disables token vision. That's a fairly clear distinction i think from Obscured Senses that should persist no? I think DMs may want to be able to easily turn off a token's vision entirely. I understand that semantically there's issues with it being called blinded because technically blind characters still have senses but since it is core foundry terminology and also exclusively a token status (not a condition) I personally think it should be left as is since it should be clear to anyone who uses Foundry regularly what it does

zithith: I'm still not happy with that to be honest. If we can't change the effect of the blinded state then we don't have it. If a GM wants to turn off a token's vision they can still do that from the token props dialog relatively quickly. I just don't think this is what Johnny et al would prefer if we can avoid it. Maybe raise it with them?

Khunkurisu: I understand Johnny made a comment about liking how Darkvision worked but not it’s name, but I do not think he has enough grasp of foundry or what players expect from it to be making the judgement call.

And for the blind thing, there’s NO reason not to have the toggle option there. We do not need to remove core foundry features players expect just to adhere to RAW. There’s no reason to enforce that on them.

Khunkurisu: We should present a custom sense for obscured vision and leave the normal base foundry vision options alone. If we show that to Johnny and he hates it, I’d rather we “wasted” time in this than failing to provide him options that include proper foundry-oriented design. There should be compromise to keep certain core features aligned with VTT expectations.

Khunkurisu: We shouldn’t be afraid to show him alternatives to his suggestions.

zithith (dismissed):

Khunkurisu (approved):
Seems good, discord discussion resolved all pending issues.

PR #331, opened 2025-04-27
fix(actor document): move active effect application to post data preparation

### LINKED ISSUE #330: Active Effects unable to use derived values from actor roll data in formula
**Describe the bug**  
Active Effects support formulas, in which actor roll data can be used. However derived fields can't be used (either null or whatever initial value is set). This is due to the fact that Foundry applies active effects between `prepareBaseData` and `prepareDerivedData` on the Actor.

**To Reproduce**  
1. Create a Character
2. Set level to 6 (tier 2)
3. Add an Item and give it a new ActiveEffect
4. Set attribute change to `system.resources.inv.max.bonus`, mode to `Add`, and value to `@tier`

**Expected behavior**  
The investiture max should become 2 (the character's tier), but remains at 1 (the default value for tier).

**Screenshots**  
_If applicable, add screenshots to help explain your problem._

**Foundry VTT Version**  
12.331

|**Name**|**Version**|
|---|---|
|**System**|0.3.0|

**Checklist:**

stanavdb:
**Type**  
- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This PR provides a workaround for the issue of accessing derived fields in Active Effect value formulas. 

**Related Issue**  
Closes #330 

**How Has This Been Tested?**  

**Checklist:**  
- [ ] My changes do not introduce any new warnings or errors.

### DIFF
--- a/src/system/data/actor/common.ts
+++ b/src/system/data/actor/common.ts
@@ -585,6 +585,8 @@ export class CommonActorDataModel<
     public prepareDerivedData(): void {
         super.prepareDerivedData();
 
+        console.log('Preparing derived data for actor', this.parent.uuid);
+
         this.senses.range.derived = awarenessToSensesRange(this.attributes.awa);
 
         // Derive defenses
--- a/src/system/data/actor/common.ts
+++ b/src/system/data/actor/common.ts
@@ -585,8 +585,6 @@ export class CommonActorDataModel<
     public prepareDerivedData(): void {
         super.prepareDerivedData();
 
-        console.log('Preparing derived data for actor', this.parent.uuid);
-
         this.senses.range.derived = awarenessToSensesRange(this.attributes.awa);
 
         // Derive defenses
--- a/src/declarations/foundry/client/data/abstract/client-document.d.ts
+++ b/src/declarations/foundry/client/data/abstract/client-document.d.ts
@@ -36,6 +36,12 @@ declare class ClientDocument {
     public prepareBaseData();
 
     /**
+     * Prepare all embedded Document instances which exist within this primary Document.
+     * @memberof ClientDocumentMixin#
+     */
+    public prepareEmbeddedDocuments();
+
+    /**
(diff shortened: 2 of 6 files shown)


src/system/data/actor/common.ts
```diff
@@ -585,6 +585,8 @@ export class CommonActorDataModel<
     public prepareDerivedData(): void {
         super.prepareDerivedData();
 
+        console.log('Preparing derived data for actor', this.parent.uuid);
```
zithith, line 588: should we swap this to a debug maybe? or potentially remove it altogether?

zithith (approved):
I'm accepting this on the grounds that I have no idea what is best to do here.
I can see now why PF2e has the Effects System in-house, this is a pretty big oversight.

I don't like the messing with the prototypes, but we don't have other options and not enough resource to do a full refactor of our own equivalent before 1.0. And as much as I'd normally be one for removing comments, I think in this case having reference to the alternatives may well be more helpful.

If you want to change the log line I'll re-approve, but on the grounds I have no better solution, happy to go forward

src/system/data/actor/common.ts
stanavdb: Nah just needs to be removed altogether. Just a left over from testing.

stanavdb:
> I'm accepting this on the grounds that I have no idea what is best to do here. I can see now why PF2e has the Effects System in-house, this is a pretty big oversight.
> 
> I don't like the messing with the prototypes, but we don't have other options and not enough resource to do a full refactor of our own equivalent before 1.0. And as much as I'd normally be one for removing comments, I think in this case having reference to the alternatives may well be more helpful.
> 
> If you want to change the log line I'll re-approve, but on the grounds I have no better solution, happy to go forward

The comment was just left over since I wasn't entirely sure which solution to go for, but I think the current solution is better. Removed the comment and marking as read for review

zithith (approved):

PR #344, opened 2025-05-02
feat(items): expand data model and usage workflow to support multiple consumption options

### LINKED ISSUE #262: Add "OR" Option to Resource Cost Field
There is room in the design for certain abilities to let you choose between two different resource costs, such as spending 1 investiture or 1 focus.

Add the ability to OR two resources, and allow the popup to let you select which to make use of.

There's really no clean way to handle this without making duplicate abilities.

**Version**  
0.3.0

**Additional context**  
N/A

**Checklist:**

NoahJGersh:
**Type**  

- [ ] Bug fix
- [x] Feature
- [x] Refactor

**Description**  
Extends the existing data model to represent resource consumption as a list. Additionally reconfigures the item sheet to support the addition of further consumption options (and their removal, by setting the type to None). Also updates the resource consumption dialog to allow the user to deselect any number of resources.

Duplicate resource types are fully supported, and handled appropriately. This is largely so users can choose to add additional costs, like additional Focus they might want to spend conditionally, as they desire.

**Data Changes**
`CosmereItem.system.activation.consume` - **Breaking**
- Old type: `ItemConsumeData`
- New type: `ItemConsumeData[]`

**Data migration is necessary.** 

**To-Do**
- [ ] Migration
- [ ] Style pass on the item activation data component and item-consume dialog

**Related Issue**  
Closes #262 

**How Has This Been Tested?**  
1. Create an item
2. Set the initial consume type to focus, with any non-zero value
3. Click the `+` to add another consumption option
4. Set the new consumption option to investiture, with any non-zero value
5. Roll the item, testing all four permutations of the resource consumption selection
6. Change the investiture option to _also_ be focus
7. Repeat step 5
8. Change either option to None- confirming that it disappears from the sheet.

To-do (style pass pending)

**Checklist:**  
- [x] I have tested my changes on Foundry VTT version: 12.311.

**Additional context**  
I built this off of the 0.3.1 branch, but it's best to leave it for 1.0.0 as a result of the breaking change. I'm only using 0.3.1 as the base branch to simplify the commit history. We should hold this PR, even if complete, until 0.3.1 is out and merged into the 1.0.0 working branch.

### DIFF
--- a/src/style/sheets/item/module.scss
+++ b/src/style/sheets/item/module.scss
@@ -316,7 +316,7 @@
             min-height: 32px;
 
             > app-id-input,
-            > .form-fields {
+            .form-fields {
                 display: flex;
                 flex-direction: row;
                 flex: 2;
--- a/src/system/applications/item/base.ts
+++ b/src/system/applications/item/base.ts
@@ -19,6 +19,7 @@ import { ComponentHandlebarsApplicationMixin } from '@system/applications/compon
 import { TabsApplicationMixin } from '@system/applications/mixins';
 import { DescriptionItemData } from '@src/system/data/item/mixins/description';
 import { ItemConsumeData } from '@src/system/data/item/mixins/activatable';
+import { SYSTEM_ID } from '@src/system/constants';
 
 const { ItemSheetV2 } = foundry.applications.sheets;
 
@@ -433,7 +434,7 @@ export class BaseItemSheet extends TabsApplicationMixin(
     }
 
     /**
-     * Helper to track activation consumption changes
+     * Helper to manage activation consumption changes
      */
     private getUpdatedConsumption(
         formData: FormDataExtended,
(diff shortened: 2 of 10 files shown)


NoahJGersh:
This is complete now, but keeping as a draft until 0.3.1 ships out.

zithith:
I did notice the edit/configure deflect dialog (where you set bonus and natural deflect) needs you to press update twice sometimes.
Not sure if that's something introduced here or not?

MangoFVTT (changes requested):
I've tested this fairly thoroughly while doing the styling and functionality all seems to work as expected. Just gave it another quick pass to make sure everything works on that front.

There is one thing to fix which is that actions no longer show anything in the cost column if there is any consume, even if there is only 1 consume. We probably want something placeholder there for now (probably just take the suggestions from the ongoing vote in discord?)

image

NoahJGersh:
> There is one thing to fix which is that actions no longer show anything in the cost column if there is any consume, even if there is only 1 consume. We probably want something placeholder there for now (probably just take the suggestions from the ongoing vote in discord?)

I caught this while working on the variable cost branch, so I'll patch that fix in

NoahJGersh:
This fix is temporary, it works but only by preserving the legacy functionality. A more thorough fix, which accounts for multiple costs being present (as well as cost ranges), will be part of the follow-up variable cost PR after this is merged in.

MangoFVTT (approved):

PR #349, opened 2025-05-06
fix(configure resources config): fix issue causing bonus value to double each time the dialog is saved

### LINKED ISSUE #346: Custom Max Resource Values Infinitely Increase with Bonus Value Assigned
**Describe the bug**  
When setting a custom maximum for a resource while an Active Effect has applied a "bonus" to the maximum (using the appropriate key of `resources.x.max.bonus`), the bonus will be included in the entered value, causing an infinitely scaling maximum resource each time the window is opened.

**To Reproduce**  
1. Create an active effect that modifies the `max.bonus` value of a resource and toggle it on
2. Open the config for that same resource
3. Set it to custom
4. Notice it shows the final value, rather than the base
5. Try to adjust the value
6. Witness the bonus increase the max twice over

**Expected behavior**  
Custom max values for a resource should use the `max.derived` value so that bonuses are not impacting its final result, thus preventing infinite stacking.

**Screenshots**  
N/A

**Foundry VTT Version**  
12.331

|**Name**|**Version**|
|---|---|
|**System**|0.3.0|

**Browser**  
Local Foundry install

**Additional context**  
N/A

**Checklist:**

stanavdb:
**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This PR fixes an issue with the configure resource dialog that causes the bonus value to be doubled each time the dialog is saved. Also resolves an issue causing the override (custom) input to re-apply the bonus after any change.

**Related Issue**  
Closes #346 

**How Has This Been Tested?**  
1. Create an active effect with: key system.resources.foc.max.bonus, mode: Add, value: 2
2. Toggle effect on
3. Open the resource config dialog for focus
4. Set it to mode custom
5. Adjust the value twice
6. Click update (value should be whatever you set custom to)
7. Open the config again
8. Click update (value should not change)

**Checklist:**  

### DIFF
--- a/src/system/applications/actor/dialogs/configure-resource.ts
+++ b/src/system/applications/actor/dialogs/configure-resource.ts
@@ -88,7 +88,12 @@ export class ConfigureResourceDialog extends HandlebarsApplicationMixin(
 
     private static onUpdateResource(this: ConfigureResourceDialog) {
         void this.actor.update({
-            [`system.resources.${this.resourceId}`]: this.resourceData,
+            [`system.resources.${this.resourceId}`]: {
+                max: {
+                    mode: this.mode,
+                    override: this.resourceData.max.override,
+                },
+            },
         });
         void this.close();
     }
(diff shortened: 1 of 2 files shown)


MangoFVTT (approved):

Khunkurisu (changes requested):
Focus is now stuck permanently on "Custom" mode. Set to derived and then cause an update on the sheet to see it reset to custom.

zithith (approved):

Khunkurisu (approved):
Regression resolved, changes now work entirely as intended.

PR #363, opened 2025-05-20
fix(skill mods): Ensured Attribute mods were properly accounted for

### LINKED ISSUE #356: Derived Data Issues
This is a list of bugs that involve derived data, modifying it and displaying it and errors that arrise with it.

First issue - Modifying Attributes via active effects do not display related skill increases and clicking on a skill rolls it with no increased values; however if you use a weapon or action that "calls" the skill test then it does do the math correctly. 
**To Reproduce**  

Image

Image

Using the following AE 

Image
This works

Image
but this stays the same

Image

However if you roll an attack

Image

**Foundry VTT Version**  
12

0.3.1

**Additional context**  
Add similar issues as comments below so they can be worked on.

**Checklist:**

zithith:
**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This change ensures that the skill displays, roll dialogs and roll calculations all correctly account for any attribute bonuses applied through Active Effects.
This is achieved mainly through the moving of where the skill mod derived data field is calculated, but also required some updates to the roll configuration dialogs to ensure correct display there too.

**Related Issue**  
Closes #356

**How Has This Been Tested?**  
Following the reproduction steps in the linked issue (see images) and also by rolling skills from the sheet.

image
image
image
image

**Checklist:**  

### DIFF
--- a/src/system/data/actor/common.ts
+++ b/src/system/data/actor/common.ts
@@ -635,26 +635,26 @@ export class CommonActorDataModel<
             this.defenses[group].derived = 10 + attrsSum;
         });
 
-        // Derive skill modifiers
-        (Object.keys(this.skills) as Skill[]).forEach((skill) => {
-            // Get the skill config
-            const skillConfig = CONFIG.COSMERE.skills[skill];
+        // // Derive skill modifiers
+        // (Object.keys(this.skills) as Skill[]).forEach((skill) => {
+        //     // Get the skill config
+        //     const skillConfig = CONFIG.COSMERE.skills[skill];
 
-            // Get the attribute associated with this skill
-            const attributeId = skillConfig.attribute;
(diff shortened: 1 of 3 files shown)


src/system/data/actor/common.ts
NoahJGersh, line 1: The comments here could use a cleanup now that the derivation has been moved in full.

NoahJGersh (commented):
I toyed around with this a bit and it's working as expected. Aside from the comment cleanup, everything looks good to me.

src/system/data/actor/common.ts
zithith: Good spot, thanks!
Removed.

NoahJGersh (approved):

PR #398, opened 2025-06-02
feat(migrations): add manual migration support for world data and compendiums

### LINKED ISSUE #392: Support manual migration invocations
Migrations function fine automatically, but it's a bit painful when working within a major version branch, and dealing with data model changes. The current "solution" to that is manually setting the version in the console, and reloading, but that's clunky.

Additionally, only world data is migrated; compendiums are not touched at all. While we want to continue excluding compendiums from automatic migrations, it would be useful to be able to manually migrate selected compendium(s). This would be a great boon for our data entry workflows, and any module developers who don't want to have to import everything and then re-export.

Add a static function to the migrations system to allow in-console invocations. This should support a "source" version, and optionally a list of compendiums to migrate.

Raw JSON migrations for compendiums work, but they're a pain. Handling everything programmatically, with full access to the foundry runtime, is ideal.

This also vastly improves the UX for major releases, where things might break and re-running migrations is critical. The existing method of doing so is clunky, and requires the console _anyways_, so this is a suitable and more well-documented method. Plus, you won't need to refresh foundry.

**Version**  
12.343

**Checklist:**

NoahJGersh:
**Type**  

- [ ] Bug fix
- [x] Feature
- [x] Refactor

**Description**  
This PR adds a global utility function to allow users to manually invoke migrations of their choice. Users specify the source and target versions, and may optionally specify a list of compendium IDs. The migration system now supports compendium data as a result, but does not do so during the automatic invocation when a version upgrade is detected.

**Related Issue**  
Closes #392 

**How Has This Been Tested?**  
Migration logic is as expected, but it should be confirmed that it runs in the first place
1. Create items both in the world and in a couple compendiums on 0.3.1
2. Give those items activation with a nonzero consumption, e.g. 1 focus
3. Close foundry
4. Build the latest version of the system on the 1.0.0 development branch
5. Open foundry
6. Confirm that the compendiums were NOT migrated (and open them, so you can see that the consumption no longer shows up), but that the items in the world WERE. If everything is good, there was no regression.
7. Open the console, and call the new function: `cosmereRPG.utils.invokeMigration("0.3", "1.0")`
8. Confirm that the compendiums were STILL NOT migrated, but that the items in the world are re-processed and the migration runs. **This confirms manual invocation on world data, in the standard way our automatic process works.**
9. Call the function again, including a list of the compendiums' IDs, e.g. `cosmereRPG.utils.invokeMigration("0.3", "1.0", ["world.some_compendium", "world.some_other_compendium"])`
10. Confirm that the compendiums ARE migrated, opening them to see the consumption there, and that the items in the world are NOT re-processed. **This confirms compendium migration functionality.**

**Checklist:**  
- [x] I have tested my changes on Foundry VTT version: 12.343.


NoahJGersh:
Some other potential testing notes:

- Confirm that running the 0.3 -> 1.0 migration on a compendium of any non-Item Document doesn't break
- Confirm that running from 0.2 -> 1.0 is fluid and functional
- Confirm that running from 0.2 -> 0.3 **while on 1.0.0** functions as expected (data is not brought all the way up to the 1.0.0 model, but should have parity with the 0.3 model).

Khunkurisu (approved):
Works as intended, huzzah. Took me a minute to realize I had to remove `Compendium.` from the id, but that's user error and can be covered by documentation/guide. I will also say that if there is a way to detect the compendium is locked *early* we should do that instead of giving the "compendium is locked" error for each item inside it.

NoahJGersh:
> I will also say that if there is a way to detect the compendium is locked *early* we should do that instead of giving the "compendium is locked" error for each item inside it.

I knew there was something I forgot. That'll be an easy fix. I'll make that then request a re-review.

Khunkurisu (approved):
The new fix for handling locked compendiums works, as does the GM check.

PR #464, opened 2025-07-11
feat: add quantity change modifiers

### LINKED ISSUE #458: Provide more options for changing quantities of items in the inventory tab
**What needs changing?**  
When trying to add/remove items in bulk from the inventory sheet, you currently have to either click the +/- controls a lot of times or go into the item to edit the quantity directly, we want there to be quicker solutions involving fewer clicks.
* Add modifier key event handlers to the +/- buttons so that when shift or ctrl is used the quantity is changed by 5 or 10/20.
* Possibly include a couple of settings so that the user can define how much they want the value to increase with each modifier key.
* It would also be great to make the number in the cell act as an input box so the value can be changed more directly too.

We could add multiple buttons (with or without modifier keys) to accomplish the same or more granularity, but the space available to the cell isn't a lot.

MangoFVTT:
**Type**  

- [ ] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  
Adds key modifiers (shift, ctrl, alt) to the quantity change buttons  so they can be increased or decreased by larger amounts than 1 (in this case, 5, 10, 50). Keybinds are fully customisable as normal.

**Related Issue**  
Closes #458

**How Has This Been Tested?**  
Tested with all  3 keybindings, both increasing and decreasing of items. Ensured currency continues to be updated normally.

**Checklist:**  
- [X] I have tested my changes on Foundry VTT version: 12.343.

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -1683,6 +1683,9 @@
         "skipDialogDefault": "Skip/Show Dialog",
         "skipDialogAdvantage": "Skip Dialog with Advantage",
         "skipDialogDisadvantage": "Skip Dialog with Disadvantage",
-        "skipDialogRaiseStakes": "Skip Dialog with Raise Stakes"
+        "skipDialogRaiseStakes": "Skip Dialog with Raise Stakes",
+        "changeQuantity5": "Increase/Decrease Quantity by 5",
+        "changeQuantity10": "Increase/Decrease Quantity by 10",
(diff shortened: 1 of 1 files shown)


stanavdb:
Question: Should we make this only partially address #458 and also make it possible to set the quantity directly from the character sheet through an input?

stanavdb (approved):

PR #506, opened 2025-07-26
fix(data): fix issue with expertises field causing it to lose track of ids during partial updates

### LINKED ISSUE #505: Manually updating expertises after using the Erudition macro breaks the character data
**Describe the bug**  
When you manually add or remove any expertise after using the Erudition macro and error occurs and the character data gets corrupted. Once this happens to a character, you can't update them anymore and they become inaccessible altogether after a reload.

This is not unique to the Erudition macro, but that is the most clear example of it. This occurs when an expertise gets set that does not define a value for the `label` property. The repro steps below set the expertise in the same way as the Erudition macro.

**To Reproduce**  
1. Create a character actor
2. Run in console: `game.actors.get('<id>').update({ 'system.expertises.armor:chain': { id: 'chain', type: 'armor' } })`
3. Open edit expertises dialog
4. Remove or add any expertise (except chain)
5. See error in console

**Expected behavior**  
Expertises data should not break when label field doesn't get set

**Foundry VTT Version**  
12.343

|**Name**|**Version**|
|---|---|
|**System**|1.0.0|
|Stormlight Handbook|1.0.1|

**Checklist:**

stanavdb:
**Type**  
- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This PR resolves the issues with updating the expertises field, causing character data to become invalid. This issue was twofold:
Firstly the expertises dialog uses the `getObjectChanges` util function to get a changes object to apply. This util function was bugged causing it to count unchanged fields with the value `undefined` as removed. This would make it return a change to remove an expertise's label if the label was `undefined` instead of explicitly set to `null`.
Secondly the `ExpertisesField` would assign its composite key in `_cleanType` even if the fields necessary to construct the composite key were missing. This would cause it to then construct a `undefined:undefined` key, which would resolve to `armor:undefined` as armor is the initial value of the expertise type. This second part is what actually breaks the data.

What would happen is: An expertise would get set with no label (`undefined`). An update would occur in the expertises dialog, which would use to `getObjectChanges` util to find the changes. The `getObjectChanges` function would return a change to remove the label of the set expertise `{ "armor:chain": { "-=label":  undefined } }`. As part of the data preparation of the update the `ExpertisesField` would attempt to clean the input data in `_cleanType`. While cleaning the input data it tries to re-assign the key. To re-assign to key it executes the static `Expertise.getKey` function using the input data (`{ "-=label": undefined }`). The `Expertise.getKey` would try to construct the composite key from the `id` and `type` field, thus returning `"undefined:undefined"`.

**Related Issue**  
Closes #505 

**How Has This Been Tested?**  
1. Create a character actor
2. Run in console: `game.actors.get('<id>').update({ 'system.expertises.armor:chain': { id: 'chain', type: 'armor' } })`
3. Open edit expertises dialog
4. Remove or add any expertise (except chain)

**Checklist:**  
- [x] I have tested my changes on Foundry VTT version: 12.343


MangoFVTT (approved):
The breaking issue that was discovered seems to be fixed, I noticed no other issues while testing

PR #508, opened 2025-07-26
fix(base actor sheet): make form events only update the fields related to the source of the event

### LINKED ISSUE #507: Updating an actor while any of its attributes are being modified by an active effect causes that value to increase
**Describe the bug**  
If you update certain fields on an actor while any of its attributes are being modified by an active effect (such as when they are Enhanced), causes those attributes to increase by the bonus applied in the AE. 

The fields effected are:
- name
- any attribute
- level
- obstacle
- purpose
- any resource value

**To Reproduce**  
1. Create a character
2. Add an effect
3. Set attribute key to `system.attributes.spd.bonus`
4. Set mode to `Add`
5. Set effect value to `1`
6. Submit effect changes
7. Update any of the above listed fields

**Expected behavior**  
Attributes so stay static

**Foundry VTT Version**  
12.343

|**Name**|**Version**|
|---|---|
|**System**|1.0.0|

**Checklist:**

stanavdb:
**Type**  
- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This PR adjusts the base actor sheet so that form events no longer cause it to update all editable fields on the sheet, but only the field that is actually being modified.
This resolves #507 because the attribute values are no longer getting updated along with the actual changes being made.

**Related Issue**  
Closes #507 

**How Has This Been Tested?**  
1. Create a character
2. Add an effect
3. Set attribute key to `system.attributes.spd.bonus`
4. Set mode to `Add`
5. Set effect value to `1`
6. Submit effect changes
7. Update the character's name
8. Ensure the speed does not go up

**Checklist:**  
- [x] I have tested my changes on Foundry VTT version: 12.343


MangoFVTT (approved):
Tested with the steps provided, seems good

PR #513, opened 2025-07-28
Translation into Catalan

NeoMahler:
**Type**  

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Other (please describe): translation

**Description**  
This adds the ca.json file with the full system translation into the Catalan language. I was told in the Discord server (username: `sparkyondo`) that the language needs to be added in another file too, but that it was fine if I just submitted the translation itself.

**Related Issue**  
I don't think there's an open issue for this translation.

**How Has This Been Tested?**  
I don't have local Foundry and my subscription to Forge VTT has expired (until I renew it in a month or so when I begin my campaign). It just JSON strings though and I've been careful with syntax errors.

**Checklist:**  

**Additional context**  
There isn't an official translation of the game into Catalan, and the Stormlight books are scheduled for translation but they aren't out either. So when it comes to both RPG and Cosmere terms, I've done my best to guess what the translator will do by comparing it to the Spanish and French translations. I'll be happy to update this file if the game gets an official translation.

I've also left ft. as feet, even though the official Spanish translation uses meters and therefore it's a safe bet a potential Catalan translation would too. But since the system does calculations with those numbers, just translating ft as meter would probably not work.

Let me know if anything else needs updating!

### DIFF
--- a/src/lang/ca.json
+++ b/src/lang/ca.json
@@ -33,7 +33,7 @@
                 "Role": {
                     "Minion": "Sequaç",
                     "Rival": "Rival",
-                    "Boss": "Amo"
+                    "Boss": "Capitost"
                 }
             },
             "Actions": {
@@ -125,7 +125,7 @@
                 "LiftingCapactiy": "Capacitat d'aixecament",
                 "RecoveryDie": "Recuperació",
                 "Deflect": "Desviament",
-                "NoObscuredSense": "No afectat per sentits enfosquits.",
+                "NoObscuredSense": "No afectat pels sentits enfosquits.",
                 "Immunities": "Immunitats"
             },
             "Deflect": {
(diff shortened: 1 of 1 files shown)


MangoFVTT (changes requested):
@NeoMahler just a note: I am gonna block this with request changes for now because we have decided that we need to first discuss with Brotherwise about how we handle fan translations into the system, and that will happen in our next check-in meeting with them which is a couple of weeks away. 

Mind you, even if we don't end up putting it into the system we'll figure out some kinda 3rd party module for fan translations, don't fear! Your translation efforts are greatly appreciated

src/lang/ca.json
```diff
@@ -0,0 +1,1712 @@
...
+                        "Edit": "Edita el prerequisit",
+                        "Delete": "Elimina el prerequisit",
+                        "TalentId": "Identificador del talent",
+                        "TalentIdPlaceholder": "Identificador del talent",
+                        "Attribute": "Atribut",
+                        "AttributeValue": "Valor",
+                        "AttributeValuePlaceholder": "Valor mínim de l'atribut",
+                        "Skill": "Habilitat",
+                        "SkillRank": "Grau",
+                        "SkillRankPlaceholder": Grau mínim de l'habilitat",
```
gxolin, line 1172: missing `"` here
```suggestion
                        "SkillRankPlaceholder": "Grau mínim de l'habilitat",
```

stanavdb:
@NeoMahler The team has decided we're going to handle community language support through modules; The core system will only come with English and any additional languages will need to be installed through modules. This is to avoid the maintenance pains that come with having a number of languages directly in system.

I'm closing this PR. If you have any questions on how to convert these changes to a module feel free to ask.

PR #609, opened 2025-11-13
Fix: Made Journal Entry Headers Editable

### LINKED ISSUE #565: system formated Journals can not have their title edited
**Describe the bug**  
When a journal is imported into a world that is formatted for the cosmere system themes its name can not be edited.

**To Reproduce**  
1. import a journal (say from an adventure)
2. try and open a journal and change its name
3. you wont be able to
4. create a new journal in world without formatting
5. notice it can be easily changed

**Expected behavior**  
all journals should be able to edit titles

**Screenshots**  
Uneditable

Image

Editable

Image

Possible Fix:
Allowing a right click menu option to be added to edit the title. 

**Foundry VTT Version**  
12.343

**Checklist:**

zithith:
**Type**  
- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
The header section of any journal entry you are looking at that is editable will now allow re-naming inline.
I've spent a fair while trying a few things and I think this is about as good as we can get for an inline option seeing as HTML <input> doesn't really do resizing and so easy options here are going to be limited.
Other benefits though are that the structure of those headers is considerably more streamlined (and I think it could be pushed further to handle the divider)

**Related Issue**  
Partially Closes #565 
(will need to repeat this on the premium module repos if we go this way)

**How Has This Been Tested?**  
* Open the Starter Pack compendium and open one of the journals from there
* Confirm that the display hasn't broken and is as expected
* Import the journal into your world
* open the imported version
* see that the title is now interactable
* edit the title
* move focus out of the journal header
* observe that the title of the journal in your side-panel has updated

... To Come ...

**Checklist:**  
- [x] I have tested my changes on Foundry VTT version: 13.356.

**Additional context**  
Potentially we might need to look at the option to add a rename button to the side-panel as in the issue, because currently there is no easy way to resize an input control, therefore anything with a particularly long title will get cut off and short ones will leave whitespace between the text and the artwork (which will differ from the non-edit version)

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -1490,6 +1490,11 @@
         },
         "EditEventRule": {
             "Title": "Edit Event Rule"
+        },
+        "RenameDirectoryEntry": {
+            "Title": "Rename {type}",
+            "Label": "New Name",
+            "Hint": "Enter a new name for this {type}."
         }
     },
     "COMPONENT": {
@@ -1558,6 +1563,7 @@
         "Description": "Description",
         "Talent": "Talent",
         "Goal": "Goal",
+        "Rename": "Rename",
         "Button": {
             "Roll": "Roll",
             "Continue": "Continue",
--- a/src/style/starter-rules/journal-common.scss
+++ b/src/style/starter-rules/journal-common.scss
@@ -75,7 +75,7 @@
             padding-left: 0;
             padding-right: 0;
 
-            p, input {
+            p {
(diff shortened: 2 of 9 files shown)


Doc-Sun:
image
Function to change works,
However as per the image, the edit area is very much cut off so you cant read it. and the font is bright white and not using the standard header colors.
image
image

Editable area does not scale the side symbols to change the font smaller or larger

PR #628, opened 2025-12-22
Made action and equipment sections on actor sheets collapsible

### LINKED ISSUE #541: Collapsible Categories in Actions and Equipment Tabs
Basically title. Being able to collapse those would be super handy cause sometimes you really just don't need to look at your Basic Actions but having them on the sheet is nice.

Uhh... collapse them.

No

**Version**  
1.0.1

**Additional context**  
N/A

**Checklist:**

posadist-revolution:
**Type**  

- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
- Made action and equipment sections on actor sheets collapsible, using precedent set with individual collapsible items
- Moved generic `collapsible header` class setup previously in `base.ts` to `notes-field.ts` to prevent interfering with the new `actions-list.hbs` collapsible header
- Made `collapsible_content` class in actor sheets only respect `expanded` of direct parents, instead of all parents
- Added new `expandSectionsByDefault` setting to configure default expand/contract of list sections

**Related Issue**  
Addresses issue #541

**How Has This Been Tested?**  
- Created new test character with a variety of different action and equipment sections, and collapsed and expanded all of those sections.
- Tested collapsing other collapsible elements (e.g. expertises section) to verify no other collapsible elements were affected by the change
- Tested enabling edit mode, collapsing or expanding elements which are only visible in edit mode, and disabling edit mode, to verify that the system can correctly handle those elements appearing and being hidden
- Tested enabling and disabling new "Expand Actor Sheet Sections By Default" setting and opening a character sheet after to verify changes were reflected

Test_Character_With_Collapsing_Actions_Sections
Test_Character_With_Collapsing_Equipment_Sections
Zealot_With_Collapsing_Actions

**Checklist:**  
- [x] I have tested my changes on Foundry VTT version 13, build 351.

### DIFF
--- a/src/system/applications/actor/components/actions-list.ts
+++ b/src/system/applications/actor/components/actions-list.ts
@@ -6,8 +6,6 @@ import {
     ItemConsumeType,
     Resource,
 } from '@system/types/cosmere';
-import { ConstructorOf } from '@system/types/utils';
-import { Talent } from '@system/types/item';
 import {
     ItemListSection,
     DynamicItemListSectionGenerator,
@@ -24,7 +22,6 @@ import AppUtils from '@system/applications/utils';
 import { AppContextMenu } from '@system/applications/utils/context-menu';
 
 // Component imports
-import { HandlebarsApplicationComponent } from '@system/applications/component-system';
 import { BaseActorSheet, BaseActorSheetRenderContext } from '../base';
 import { SortMode } from './search-bar';
 
--- a/src/system/applications/actor/components/actions-list.ts
+++ b/src/system/applications/actor/components/actions-list.ts
@@ -12,6 +12,7 @@ import {
     ItemListSection,
     DynamicItemListSectionGenerator,
 } from '@system/types/application/actor/components/item-list';
+import { ActorItemListComponent } from '@system/applications/actor/components/item-list';
 
 // Documents
 import { CosmereItem } from '@system/documents/item';
@@ -316,11 +317,7 @@ const MISC_SECTION: ItemListSection = {
     filter: () => false, // Filter function is not used for this section
 };
 
-export class ActorActionsListComponent extends HandlebarsApplicationComponent<// typeof BaseActorSheet
-// TODO: Resolve typing issues
-// NOTE: Use any as workaround for foundry-vtt-types issues
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-any> {
+export class ActorActionsListComponent extends ActorItemListComponent {
     static TEMPLATE = `systems/${SYSTEM_ID}/templates/${TEMPLATES.ACTOR_BASE_ACTIONS_LIST}`;
 
     /**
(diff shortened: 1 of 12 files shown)


src/system/applications/actor/components/adversary/actions-list.ts
```diff
@@ -55,6 +55,15 @@ export class AdversaryActionsListComponent extends ActorActionsListComponent {
         const searchText = context.actionsSearch?.text ?? '';
         const sortMode = context.actionsSearch?.sort ?? SortMode.Alphabetic;
 
+        // Ensure all sections have an expand state record defaulting to true
+        this.sections.forEach((section) => {
+            if (!(section.id in this.sectionState)) {
+                this.sectionState[section.id] = {
+                    expanded: true,
```
Khunkurisu: Rather than defaulting to true, we should use a setting to determine the default. The setting can default to true. See the item description auto-expansion setting for implementation.

Khunkurisu (changes requested):

Khunkurisu:
> Would it be valuable to add a setting for whether item descriptions on actor sheets should expand their descriptions by default in a similar way?

This is probably a good idea, but it should go in a separate PR.

stanavdb (changes requested):
Thanks for the PR! Not taken a look at the code yet, but there's a regression on the styling: 
**release-2.0.4**
image

**PR**
image

Also, to be more consistent with how the headers behave we should probably use a chevron here. For reference:
image

src/system/applications/actor/components/actions-list.ts
```diff
@@ -333,13 +338,50 @@ any> {
...
+        sectionElement
+            .find('a[data-action="toggle-section-collapsed"')
+            .empty()
+            .append(
+                this.sectionState[sectionId].expanded
+                    ? '<i class="fa-solid fa-compress"></i>'
+                    : '<i class="fa-solid fa-expand"></i>',
+            );
+    }
+
```
stanavdb: This is adding quite a lot of duplicate code, between this, adversary actions list, the equipment list, and (a tiny bit) the notes field. Can we generalise the collapsible logic for all applications? Maybe as a mixin?

src/templates/actors/components/actions-list.hbs
```diff
@@ -13,6 +14,20 @@
...
+                    data-tooltip="GENERIC.HideDetails"
+                    {{else}}
+                    data-tooltip="GENERIC.ViewDetails"
+                    {{/if}}
+                >
+                    {{#if sectionState.expanded}}
+                        <i class="fa-solid fa-compress"></i>
+                    {{else}}
+                        <i class="fa-solid fa-expand"></i>
+                    {{/if}}
```
stanavdb: Not entirely specific to this PR but, since the application/component isn't re-rendered when a section is collapsed, the tooltip doesn't get updated. The toggle handler only updates the icon. 
This also happens when expanding/collapsing the items themselves, so this PR matches the current (bugged) behaviour.

stanavdb (changes requested):

src/system/applications/actor/components/actions-list.ts
posadist-revolution: I'll look into this! I'm still learning a lot of stuff with CSS/TypeScript because I haven't used them before, so thanks for the tip on looking into mixins for this! 😄

posadist-revolution: I took a look at how the mixin pattern is used both generally and in this codebase specifically, and kind of came to the conclusion that with where I'm at in my TypeScript/Handlebars/Foundry API understanding, fully generalizing collapsible logic for all applications would be too big of a lift for me to handle right now. I definitely agree with trying to limit the duplicate code between all of these different files- to that end, I moved a bunch of the duplicate code into a new `item-list.ts` file under a new `ActorItemListComponent`, which is extended by both `ActorActionListComponent` and `ActorEquipmentListComponent`. Does that resolve the duplicate code issues enough for this PR?

src/system/applications/actor/components/actions-list.ts
```diff
@@ -435,6 +356,18 @@ any> {
             sortMode,
         );
 
+        // Ensure all sections have an expand state record defaulting to settings.expandSectionByDefault
+        const expandSectionDefaultSetting =
+            game.settings?.get('cosmere-rpg', 'expandSectionsByDefault') ??
```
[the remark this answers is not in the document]
posadist-revolution: Fixed

src/system/applications/actor/components/equipment-list.ts
```diff
@@ -266,6 +187,17 @@ any> {
             this.prepareSection(ItemType.Loot),
         ];
 
+        // Ensure all sections have an expand state record defaulting to settings.expandSectionByDefault
+        const expandSectionDefaultSetting: boolean =
+            getSystemSetting(SETTINGS.SHEET_EXPAND_SECTIONS_DEFAULT) ?? true;
```
Khunkurisu: Since this line applying a default state is repeated in both list class' `_prepareContext()` method, assigning a getter in the item-list parent class would likely be a cleaner solution. Minor change, but there's a good bit of repeated code here *in general* that can be split off.

src/system/applications/actor/components/equipment-list.ts
```diff
@@ -266,6 +187,17 @@ any> {
...
+        this.sections.forEach((section) => {
```
Khunkurisu: This loop is another bit of repeated code in both child classes.

Khunkurisu (changes requested):
Basically just a bit more consolidation since you decided to make that part of this PR (not complaining), which needs to be done so it's not halfway. Besides that, just merge 2.1.0 into this once you're done so you're up to date and it'll be ready to go.

src/system/applications/actor/components/equipment-list.ts
posadist-revolution: Done!

Khunkurisu (approved):

PR #630, opened 2025-12-23
Improved prototype token default settings

### LINKED ISSUE #278: Set Up Default Token Configuration by System
Currently the system doesn't really initialize tokens for characters or adversaries when creating the relevant actor sheet.

Ideally a decent number of the token values should be set to some pretty basic defaults. Friendly disposition for characters, resource bars for health and focus, etc.

Most of this can be set up by a GM using the default token config menu, but it's per-world and having system defaults pre-configured for people is always nice.

**Version**  
0.3.0

**Additional context**  
N/A

**Checklist:**

posadist-revolution:
**Type**  

- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
1. Adds default actor sheet flag setup for `automation.token.size\sight` and `sheet.hideUnranked`.
2. Sets prototype tokens on actor create to default to:
- `displayBars`: display when hovered by owner
- `bar1`: Track one actor resource from `cosmere-rpg.defaultTokenBar1Value`
- `bar2`: Track one actor resource from `cosmere-rpg.defaultTokenBar2Value`
- `width`: Token width in grid spaces generated from `actor.system.size`
- `height`: Token height in grid spaces generated from `actor.system.size`
- `sight.range`: Token vision radius in darkness generated from `actor.system.senses.range`
- `sight.visionMode`: Token vision type set to `sense`

3. Adds a hook in src/system/hooks/actor.ts to modify prototype token settings when corresponding actor values have changed, gated behind new actor flags `automation.token.size` and `automation.token.sight` which default to a global setting for `defaultTokenAutomationSight\Size` when an actor is created.

**Related Issue**  
Resolves #278 

**How Has This Been Tested?**  
- Created a new character and verified that displaybars settings were correct
- Imported a Chull (large) actor sheet and verified that modifying the size to "huge" caused prototype tokens placed to have a grid size of 3 units, and modified it again to reset it to "large" and saw that the prototype tokens had a grid size of 2 units.
- Pulled a Chasmfiend from a compendium to a map and verified that the display included both health and focus bars, displayed when hovered by an owner.
- Created a new character and verified that senses settings were correct
- Adjusted new character senses using the "Configure Senses Range: Custom" dialog (setting it to 40, as well as setting "Ignores Obstructed" and confirmed changes were reflected in prototype token
- Adjusted new character senses by adjusting character Awareness and confirmed changes were reflected in prototype token
- Adjusted new character senses by adjusting character Awareness all the way to 10 and confirmed unlimited senses range changes were reflected in prototype token
- Created multiple copies of an actor on a canvas, modified the actor size and senses range via the actor sheet, and verified that changes were reflected in all actor tokens

**Checklist:**  
- [x] I have tested my changes on Foundry VTT version 13 build 351.


src/system/documents/actor.ts
```diff
@@ -314,7 +314,79 @@ export class CosmereActor<
...
+            case Size.Large:
+                prototypeTokenSize = 2;
+                break;
+            case Size.Huge:
+                prototypeTokenSize = 3;
+                break;
+            case Size.Garguantuan:
+                prototypeTokenSize = 4;
+                break;
+        }
```
stanavdb: To make this easier to expand/configure later, please move the size mapping over to the `CONFIG`. Also token size should probably default to `1`, in case the size can't be mapped for some reason.

src/system/documents/actor.ts
```diff
@@ -314,7 +314,79 @@ export class CosmereActor<
...
+        const flags = {
+            'cosmere-rpg': {
+                sheet: {
+                    autosetPrototypeTokenValues: true,
+                    hideUnranked: true,
+                },
+            },
+        };
+
+        this.updateSource({ flags });
```
stanavdb: Only need to update source once:
```ts
this.updateSource({ prototypeToken, flags });
```

src/system/utils/handlebars/index.ts
```diff
@@ -107,7 +107,7 @@ Handlebars.registerHelper(
 
 Handlebars.registerHelper(
     'isNumMax',
-    (value: number) => value === Number.MAX_VALUE,
+    (value: number) => value === Number.MAX_SAFE_INTEGER,
```
stanavdb: For clarity, can we rename this helper to `isNumMaxSafeInt` / `isNumMaxSafeInteger`?

src/system/hooks/actor.ts
```diff
@@ -194,6 +194,105 @@ Hooks.on('updateActor', (actor: CosmereActor, changed: Actor.UpdateData) => {
...
+        }
+
+        // Send prototype token changes
+        if (Object.keys(prototypeTokenChanges).length > 0) {
+            void actor.prototypeToken.update(prototypeTokenChanges);
+        }
+    }
+});
+
+Hooks.on('updateActor', (actor: CosmereActor, changed: Actor.UpdateData) => {
```
stanavdb: When hooking into `updateActor`, be aware that its executed for all clients. We don't want all clients to try to update the prototype token (most won't even have permission and would get errors), so you have to check if the local user is the one who triggered the update. (or move all logic over the `preUpdateActor`, since that runs client side only)

stanavdb (changes requested):
Thanks for the PR! Something that's not addressed in this implementation yet, is that this should all be configurable through settings. 
As a GM I should be able to choose whether I want to automatically configure the bars (and probably also how), size & vision (or any combination).

src/system/documents/actor.ts
```diff
@@ -314,7 +314,79 @@ export class CosmereActor<
...
+
+        // Configure default actor flags
+        const flags = {
+            'cosmere-rpg': {
+                sheet: {
+                    autosetPrototypeTokenValues: true,
+                    hideUnranked: true,
+                },
+            },
+        };
```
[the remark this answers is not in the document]
posadist-revolution: These automation flags have been split out now, matching your suggestion!

src/system/hooks/actor.ts
```diff
@@ -194,6 +194,105 @@ Hooks.on('updateActor', (actor: CosmereActor, changed: Actor.UpdateData) => {
...
+            foundry.utils.mergeObject(prototypeTokenChanges, {
+                width: prototypeTokenSize,
+                height: prototypeTokenSize,
+            });
+        }
+
+        // Senses changes
+        if (
+            foundry.utils.hasProperty(changed, `system.senses`) ||
+            foundry.utils.hasProperty(changed, `system.attributes.awa`)
```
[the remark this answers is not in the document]
posadist-revolution: Updated to diff the senses range between the prototype token and the current value, since the prototype token if the user is using automation seems to be a fairly good store of "What the previous value was".

src/system/hooks/actor.ts
posadist-revolution: Added a check to verify the local user is the one who triggered the update before moving forward with this hook. Thanks!

src/system/hooks/actor.ts
```diff
@@ -194,6 +194,105 @@ Hooks.on('updateActor', (actor: CosmereActor, changed: Actor.UpdateData) => {
...
+                    range: sensesData?.range?.value,
+                    visionMode: 'sense',
+                },
+            });
+        }
+
+        // Send prototype token changes
+        if (Object.keys(prototypeTokenChanges).length > 0) {
+            void actor.prototypeToken.update(prototypeTokenChanges);
+        }
```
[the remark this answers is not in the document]
posadist-revolution: Added propagation of these values to all `actor.getActiveTokens` in the newest patchset

src/system/utils/handlebars/index.ts
posadist-revolution: Done

posadist-revolution:
> As a GM I should be able to choose whether I want to automatically configure the bars (and probably also how), size & vision (or any combination).  

Settings for all of these have been added in the most recent patchset- it may not be the best way to organize those settings though, and I'm definitely open to suggestions if there's a better way!

src/system/documents/actor.ts
posadist-revolution: Moved size mapping to be part of `SizeConfig` referenced through `CONFIG.COSMERE.Sizes`, and fixed overwriting for explicitly-passed token size and bar data. Thanks for the reference, it was very helpful!
Sidenote- sorry I sent out a re-review request and then submitted a bunch more commits to this, I genuinely thought I had addressed your comments and then realized I was fully atop Mount Stupid.

src/system/settings.ts
```diff
@@ -71,6 +93,73 @@ export function registerSystemSettings() {
...
+            scope: 'client',
+        },
+        {
+            name: SETTINGS.TOKEN_DEFAULT_BAR_2_VAL,
+            default: TokenBarOptions.focus,
+            scope: 'client',
+        },
+    ];
+
+    tokenOptions.forEach((option) => {
```
zithith: another here

zithith (dismissed):
Most of it looks good from what I can see.

Made some optimisation suggestions. Not really a huge deal, but would prefer to use the better option when it costs us nothing.

Once these are done and the rest of the team are happy with their comments we can approve 👍

src/system/settings.ts
posadist-revolution: Changed!

src/system/hooks/actor.ts
```diff
@@ -194,6 +195,97 @@ Hooks.on('updateActor', (actor: CosmereActor, changed: Actor.UpdateData) => {
...
+
+        // Send active token changes
+        actor.getActiveTokens(false, true).forEach((token: TokenDocument) => {
+            const activeTokenChanges = {};
+            foundry.utils.mergeObject(
+                activeTokenChanges,
+                prototypeTokenChanges,
+            );
+            void token.update(activeTokenChanges);
+        });
```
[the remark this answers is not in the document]
posadist-revolution: Changed!

src/system/settings.ts
```diff
@@ -39,6 +44,16 @@ type SystemSettingsConfig = {
...
 } & {
     [key in `${typeof SYSTEM_ID}.${typeof SETTINGS.SHEET_SKILL_INCDEC_TOGGLE}`]: boolean;
+} & {
+    [key in `${typeof SYSTEM_ID}.${typeof SETTINGS.AUTOMATION_TOKEN_FLAGS_BARS}`]: boolean;
+} & {
+    [key in `${typeof SYSTEM_ID}.${typeof SETTINGS.AUTOMATION_TOKEN_FLAGS_SIGHT}`]: boolean;
+} & {
+    [key in `${typeof SYSTEM_ID}.${typeof SETTINGS.AUTOMATION_TOKEN_FLAGS_SIZE}`]: boolean;
+} & {
+    [key in `${typeof SYSTEM_ID}.${typeof SETTINGS.TOKEN_DEFAULT_BAR_1_VAL}`]: string;
```
stanavdb: For improved typing, you can use a string literal type here (Resource can be imported from types/cosmere)
```suggestion
    [key in `${typeof SYSTEM_ID}.${typeof SETTINGS.TOKEN_DEFAULT_BAR_1_VAL}`]: `resources.${Resource}`;
```

stanavdb (dismissed):
Still need to test locally if everything's working, but have one more necessary change and a couple of small improvements.

src/system/settings.ts
posadist-revolution: Unless I'm mistaken this way doesn't allow for a "none" resource to be passed in, which I was using to allow users to set just one of the two bars to be automatically assigned- do you think I should forgo the "none" option and make the string literal change, or is there another option that I'm not seeing?

src/system/hooks/actor.ts
```diff
@@ -197,101 +198,91 @@ Hooks.on('updateActor', (actor: CosmereActor, changed: Actor.UpdateData) => {
...
+            if (Object.keys(prototypeTokenChanges).length > 0) {
+                void actor.prototypeToken.update(prototypeTokenChanges);
+
+                // Send active token changes
+                for (const token of actor.getActiveTokens(false, true)) {
+                    const activeTokenChanges = {};
+                    foundry.utils.mergeObject(
+                        activeTokenChanges,
+                        prototypeTokenChanges,
+                    );
```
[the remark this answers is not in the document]
posadist-revolution: Fixed!

src/system/documents/actor.ts
```diff
@@ -55,6 +55,7 @@ import { containsExpertise } from '@system/utils/actor';
 import { SYSTEM_ID } from '@system/constants';
 import { HOOKS } from '@system/constants/hooks';
 import { AnyObject } from '@league-of-foundry-developers/foundry-vtt-types/utils';
+import { prototype } from 'events';
```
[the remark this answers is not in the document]
posadist-revolution: Fixed, thank you!

src/system/documents/actor.ts
```diff
@@ -314,79 +315,103 @@ export class CosmereActor<
...
-            case Size.Large:
-                prototypeTokenSize = 2;
-                break;
-            case Size.Huge:
-                prototypeTokenSize = 3;
-                break;
-            case Size.Garguantuan:
-                prototypeTokenSize = 4;
-                break;
+        if (game.settings.get(SYSTEM_ID, 'defaultTokenAutomationBars')) {
```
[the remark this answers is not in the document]
posadist-revolution: Fixed!

posadist-revolution:
> Haven't had a chance to review this yet but biggest thing for the size automation: be sure to set a flag (can be true by default) on the actor for whether or not to automatically set the size, and then have your code check it's true before doing any automation. We plan to add a button for toggling that flag in the future, so supporting it in this PR is ideal.

@Khunkurisu I wanted to ask more about this- I was thinking about this for when I use it with my tables, and thinking it'd be nice to have a migration to set these flags on all existing actors (so I don't have to do any nonsense with setting flags manually.) I'm thinking having a migration to force everybody to set those flags to true is maybe not great when users might have settings they already like, so maybe that isn't the best idea (unless migrations can have options the users can select) and the current settings options mostly just configure prototype token defaults when an actor is created, so any pre-existing actors with this change are locked in to whatever the settings were when you created the actor until we add that button. 
Does it make sense to add this button(s) now as part of this change, so users don't have any period where they're just locked in once an actor is created? If so, I'd love to hear more about what y'all were thinking as far as UI layout and stuff with these options on an actor sheet, and I could take a swing at it here.

src/system/settings.ts
Khunkurisu: You can add `| "none"` to allow that as a valid string literal as well, though I may be misunderstanding this question.

src/system/settings.ts
```diff
@@ -51,6 +66,13 @@ export const enum TargetingOptions {
     PrioritiseTargeted = 4,
 }
 
+export const enum TokenBarOptions {
+    health = `resources.hea`,
+    focus = `resources.foc`,
+    investiture = `resources.inv`,
+    none = ``,
+}
```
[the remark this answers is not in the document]
posadist-revolution: Resolved in most recent patchset!

posadist-revolution:
This has been rebased on 2.0.5 and should be ready to be merged whenever!

Khunkurisu (approved):
All requested changes seem to be handled, and prior reviews stale so I've given it a look and seems fine.

PR #665, opened 2026-02-17
chore(templates): update default values for issue and pr templates

Khunkurisu:
this basically just adds some automation/standardization to the current templates


.github/ISSUE_TEMPLATE/1-bug_report.yml
```diff
@@ -0,0 +1,111 @@
...
+  attributes:
+    label: Steps To Reproduce
+    description: Steps to reproduce the behavior.
+    placeholder: |
+      1. Go to '...'
+      1. Click on '...'
+      1. Scroll down to '...'
+      1. See error
+  validations:
+    required: false
```
stanavdb, line 38: Is there a good reason not to make repro steps required?

Khunkurisu: No, but since the field is just a text input w/ default value required doesn't actually *enforce* much of anything.

stanavdb (approved):

Issue #645, opened 2026-02-08
Actor portrait changes not persisting


Dez384:
**Describe the bug**  
In Cosmere RPG 2.0.4, after changing the portrait of an actor, the change does not persist and the portrait image path will revert once any change is made to the sheet or it refreshes.

**To Reproduce**  
_Steps to reproduce the behavior:_
1. Update to Cosmere RPG 2.0.4
2. Open any actor sheet (player or adversary)
3. Toggle Edit mode on
4. Click on the Portrait
5. In the filepicker choose a new image
6. See the portrait change. In the sidebar, the actor icon will not change as expected
7. Make ANY change to the character sheet (or close the sheet)
8. Behold the portrait revert to whatever its previous image was

**Expected behavior**  
Using the filepicker should change the portrait image path

**Screenshots**  
From Discord user Chaos:

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/98da11bd-d9e0-4841-9b38-817102ccdb53" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/8481acea-1227-43f2-86b2-ea00311d8058" />

**Foundry VTT Version**  
v13.351

**Cosmere RPG System & Module Versions**  
|**Name**|**Version**|
|---|---|
|**System**|_2.0.4_|

**Browser**  
Foundry App

**Additional context**  
DocumentV2 default 'data-action="editImage"' is not working for actors despite working for items.

**Checklist:**
- [x] I have searched the existing issues to ensure this bug has not already been reported

PR #666, opened 2026-02-17
Fix: Actor Portraits Not Persisting 645

Khunkurisu:
…lity

**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This resolves the issue of actor portraits not persisting by reverting to our original, custom handling of image editing and related action. In order to prevent loss of The Forge file browser support, custom handling now creates an instance of `FilePicker.implementation` instead of the base FilePicker. It also keeps the data-edit value assigned to 'img' to provide module compatibility with Tokenizer (and likely others).

**Related Issue**  
#645 

**How Has This Been Tested?**  
- Install the Forge VTT asset browser module, and the Tokenizer module
- In a world, enable Forge VTT module
- Open up an actor and toggle edit mode on
- Click the actor image and see that the file browser gains the appropriate tab(s)
- Select an image and submit
- Toggle edit mode off, notice changed image persist
- Enable Tokenizer module
- Open an actor and toggle edit mode on
- Click the actor image and see that the tokenizer window appears
- Make changes and hit apply
- See that changes apply and persist appropriately, including when edit mode is toggled off

**Checklist:**  

**Additional context**  
Note that I opted to also revert + update the item actions, even though they worked previously. This is so that we can continue to have control over them if future updates warrant it, while still providing that functionality/compatibility the original removal aimed to address.

### DIFF
--- a/src/system/applications/item/components/header.ts
+++ b/src/system/applications/item/components/header.ts
@@ -20,7 +20,7 @@ any> {
      */
     /* eslint-disable @typescript-eslint/unbound-method */
     static ACTIONS = {
-        'edit-img': this.onEditImg,
+        'edit-image': this.onEditImg,
     };
     /* eslint-enable @typescript-eslint/unbound-method */
 


Dez384 (changes requested):
Only tested on Foundry app, not The Forge VTT:
1. Changing actor portraits works. 
2. Changing item portraits does not. The data action on the item sheet doesn't match the action on the item. Making the two match will fix the issue.

PR #683, opened 2026-03-27
feat(advancement): refactor advancement system to support arbitrary overrides

### LINKED ISSUE #682: Support alternate advancement rules
The current advancement system is quite limited. Each level has a specific course of action to take, but homebrewers may want their own ancestry to have its own unique rules. Additionally, we already need to improve advancement to enable work on #309 , where we can actually track the rules that have been handled by the (eventual) character builder.

The current advancement system is a collection of disparate utility functions. These should become methods for an overall `AdvancementManager` class, which is the sole source of truth for any advancement rules or their overrides. An instance of that class should exist in a globally-accessible location, e.g. `game.system.utils.advancementManager`. It should be able to handle additional override registration, allowing for any of the current given rules to be modified in one way or another. Numeric overrides should be either absolute (complete override) or relative (add some `delta` to the default).

We should also extend the current rules to support some other, more granular changes. For now, the maximum attribute points at a given level extends to _all_ attributes at once. This also applies to skill ranks. While overriding the general maximum needs to be possible, a more subtle issue is that _individual_ attributes or skills should be able to be overridden on a case-by-case basis.

We should also develop a new set of hooks for module developers to be able to hook into the advancement process, especially to enable extensions of the character builder in the future.

We could keep things in the current state, but having a dedicated class simplifies a lot of the code and compartmentalizes things really well, especially for module developers and foundry power users. There's no way around supporting these overrides as it is.

**Additional context**  
This feature blocks #309 

**Checklist:**

NoahJGersh:
**Type**  

- [ ] Bug fix
- [x] Feature
- [x] Refactor

### **Description**  
The current advancement setup doesn't permit deviation. However, it would be very useful to homebrewers and power users to define item-based overrides to modify some way a character levels up. Some examples could be additional or fewer skill ranks gained at a level, or even the cap for a given skill being adjusted. Another possibility is modifying the amount of health gained at a level. The advancement data model needs to reflect these possibilities, especially so that the upcoming character builder can properly represent the exact steps available to a given character, and track their progress accordingly.

Advancement has been moved into its own class for override registration and handling, advancement-based data derivation, and general rule management. This can be used via the `AdvancementManager` class, which is also accessible at runtime via `game.system.utils.AdvancementManager`. The methods are all static, so every user, module, or system function referencing the known rules and overrides are guaranteed to be seeing the same ones.

### **Advancement Rules**
The advancement config has been shifted slightly. Most of its standard fields for a given level have been moved to their own object (`grants`), while any values representing stat maxima have been moved and renamed into a new `maxStats` structure.

So far, two stat maxima can be set: `attributes` and `skills`. Each one has a `base` maximum, which must be present at each level. However, other keys can be specified, e.g. `maxStats.attributes.str` represents a character's maximum possible value for strength, or `maxStats.skills.hwp` which represents the maximum possible number of skill ranks that can be allocated to Heavy Weapons.

A new `AdvancementRule` class has been implemented to conform to this data model, and provide some additional override functionality.

### **Advancement Overrides**
The core of this update is to support overrides. Now, _any_ of the `grants` or `maxStats` can be overridden, on a case by case basis.

Overrides apply to a range of `levels`; this is an inclusive range of the provided `min` and `max` levels. If `min` is left undefined, it defaults to `1`. If `max` is left undefined, it defaults to `Infinity`, so it functionally applies to all levels beyond and including the `min` level.

An override has a `type`, which is either `grants` (modifies a key in `grants`) or `maximum` (modifies a key in `maxStats`). It also has a `mode`, such as `absolute` which completely changes the value. Numeric fields can also be changed with a `relative` mode, which adds the value to what already exists on the rule.

The following values are a bit more self explanatory:
`key: string`: the actual key to change, e.g. the generic key 'health' or the max stat key 'attributes'
`stat: string`: for max stats only, this can be 'base' for the general cap, or a more specific key, e.g. 'str' or 'hwp' as explained above.
`value: number | boolean`: the actual modification to the rule.

The `AdvancementOverride`, when created, does assert that the type of value matches the correct type for the field, that max stats are only numeric, and that even the stat properly belongs to the specified key. This keeps behavior well-defined and straightforward, so that bad data will be rejected.

When a rule is overridden, the default behavior is to clone itself and mutate the clone directly with the provided override(s). This helps preserve the original state of the rule, so that overrides don't change the way character advancement works for _all_ characters, just the relevant ones. The internal self-mutating helper function, `AdvancementRule.applyOverrideInPlace`, is still exposed so that unnecessary clones aren't created when someone _does_ want to change a rule as it is.

### **The Add/Remove Override Handlers**
As a means to allow items to provide overrides to Actors, a couple of new item event system handlers have been implemented.

**"Apply Advancement Overrides"** allows the configuration of any number of overrides, and will apply them to the actor that owns the item when triggered.

**"Remove Advancement Overrides"** will remove all overrides from the owning actor when triggered, as long as they were provided by the configured `sourceId`. If left undefined, the handler will remove overrides granted by the triggered item itself.

### **Advancement Manager**
The manager maintains a list of default advancement `rules`, as well as a structure of `overrides`. The latter structure has a list of `global` overrides (applied to all characters), and an `items` mapping from an item's `system.id` to a list of overrides to apply when that item is present on the character.

To simplify override configuration, the `AdvancementManager` provides a couple of registration helpers. It initializes itself during the main system init cycle, so the default configured rules and overrides will be present before any other registration can occur. When registered, overrides are automatically sorted into the above configuration, based on their `priority` and `levels`.

Additionally, the `AdvancementManager` will prioritize override _sources_ in the following order:
1. Global overrides
    -  Source: `AdvancementManager.overrides.global`
2. Actor-specific overrides
    - These are usually produced via the item event system.
    - Source: `CosmereActor.system.advancement.overrides`
3. Registered overrides
    - Additional overrides registered by macros, modules, etc.
    - Only apply when a relevant item is found on the given Actor.
    - Source: `AdvancementManager.overrides.items.<item.system.id>`

### **Caveats and TODOs**
- Per-skill and per-attribute maximums are not fully supported by the system data model. Their existence here will allow those maxima to be derived (like how the total attribute points can be derived currently), but also serves to be forward-compatible with those potential future changes.
- There is not an elegant way to provide a complete override to a rule (rather than changing a single value). One possible solution could simply be letting users provide a rule-like structure, which internally becomes transformed into a set of `absolute` overrides. For now, though, they can create their own set of absolute overrides to achieve the same effect.
- This PR does NOT support formula inputs for numeric overrides. For now, they are strictly numeric. This system has been built with the possibility of expanding the override data types to support strings, or other data types, as necessary in the future.

### **Related Issue**  
Closes #682 

### **How Has This Been Tested?**  
1. Create a Character actor. Adversaries should not work, but you shouldn't get any errors by following these steps either.
2. Create an item (any type will work)
3. Add an event rule to the item with the following details:
  - Trigger: `Added to Actor`
  - Type: `Grant Advancement Overrides` 
4. Add at least one override to the event rule configuration.
  - A good one to test is levels 2 - 4, Grants Absolute Health, value of 10, priority of 0.
5. Add another event rule to the item with the following details:
 - Trigger: `Removed from Actor`
 - Type: `Remove Advancement Overrides`
 - Leave the source ID blank to revoke only this item's overrides.
6. Add the item to the actor
7. Set the actor's level to somewhere within the level range, or outside of it. Confirm that the values change accordingly.
8. Remove the item from the actor. Confirm that the advancement changes you configured are properly reset.
9. Adjust the override rule(s) and repeat.
10. Create an additional item with its own rules and priorities, to confirm that overrides are applied correctly. Higher priority rules should override lower priority ones. Otherwise, both should coexist without change.

For source ID configuration...
1. Get the UUID of an item, on an actor, that provided advancement overrides to said actor.
2. Create a new item with the following event rule:
  - Trigger: `Added to Actor`
  - Type: `Remove Advancement Overrides`
  - Source ID: `<the UUID of the other item>`
3. Add the new item to the actor and confirm that the overrides are now gone
4. To re-establish the old overrides without removing the old item, you can change the original source item's Grant Advancement Overrides event trigger to `Updated`, save the configuration, and then set it back to `Added to Actor`. There's room for improvement here with the UX, but it's already a strange use case and so isn't a priority for this MVP build.

For override hooks...
1. Set `CONFIG.debug.hooks = true` in the console
2. Confirm that, upon adding items to the actor, it fires the relevant hooks (`cosmere-rpg.preOverrideAdvancement`, `cosmere-rpg.overrideAdvancement`)

And finally, to test override registration...
1. In the console, call the static registration method for a global override, e.g.
```ts
game.system.utils.AdvancementManager.registerAdvancementOverride({
  type: "max-stat",
  mode: "relative",
  key: "skills",
  stat: "base",
  value: 1,
  priority: 0,
  levels: {
    min: 1,
    max: 3,
  }
})
```
2. Do the same with an item-specific override. To do this, use an item which has a system id, e.g. for an ancestry with the configured id of `'human'`:
```ts
game.system.utils.AdvancementManager.registerAdvancementOverride({
  type: "grants",
  mode: "absolute",
  key: "healthIncludeStrength",
  value: true,
  priority: 0,
  levels: {
    min: 1,
    max: 1,
  }
}, "human")
```
3. Use two actors, one with said item and one without. Confirm that they both take into account the global override, but that only the actor with that item includes the item-specific override. Note that even without priority taken into account, the item-specific override will takes precedence over overrides configured on the item directly and global overrides.
4. You can also confirm that the overrides are registered correctly by logging the value of all registered overrides, i.e.
```ts
game.system.utils.AdvancementManager.overrides
```
5. Ensure the relevant hooks fire during this process as well (`cosmere-rpg.preRegisterAdvancementOverride`, `cosmere-rpg.registerAdvancementOverride`)

### **Screenshots (if applicable)**  
image

### **Checklist:**  


NoahJGersh:
Got everything caught up (properly this time) with the 3.0.0 branch

Khunkurisu:
This looks really good and thorough at a quick glance but I’ll have to do a proper review due to the complexity. Is this strictly MVP for 3.0.0 or would it fit fine into 3.1.0?

PR #720, opened 2026-05-18
refactor(talent): remove path/ancestry/power data model fields

### LINKED ISSUE #314: Remove path/ancestry from talent data model
**What needs changing?**  
Talents used to be linked to paths through their own data model (including prerequisites). With the release of 0.3.0 this has been moved to Talent Trees. However talents still have fields in their data model that refer to its path/ancestry/power. These fields need to be removed as they no longer serve any purpose.

Image

-

mirandaguillaume:
**Type**
- [ ] Bug fix
- [ ] Feature
- [x] Refactor

**Description**
Talents used to carry their path / ancestry / power linkage directly on their data model. Since 0.3 that responsibility has moved to talent trees, but the three string fields stayed on the schema, with no consumer reading them outside their own derived-data branch.

This PR removes the dead surface:

- `TalentItemDataModel`: `path`, `ancestry`, `power` removed from the schema; `hasPath / hasAncestry / hasPower` derived flags and the `prepareDerivedData` branch that filled them are removed.
- `TalentItemSheet`: the `onFormEvent` override that coerced an empty `system.path` to `null` is gone, along with the `isPathTalent / isAncestryTalent / isPowerTalent` context flags (no template branch reads them anymore). The `form: { handler }` block in `DEFAULT_OPTIONS` is dropped so the sheet now inherits the base handler, matching the pattern used by `AncestrySheet`.
- `talent-details-tab.hbs`: the three `{{#if isXTalent}}` sections are removed. The `system.specialty` input nested in the path block was already dead (no matching schema field anywhere in the codebase) and is removed in the same pass — flagging it for visibility.
- `actions-list.ts`: the `paths` section was seeding `system.path = path.system.id` when creating a new talent from an actor's path. The filter on this section already uses `isRelatedTo(path, Parent)`, so the seeding has been writing to a now-dead key. The line is removed; new talents created from this UI are no longer auto-linked and the user attaches them via a talent tree.
- `lang/en.json`: drops the nine unused `COSMERE.Item.Sheet.Talent.{PathId,SpecialtyId,AncestryId}{,Placeholder,Description}` keys.

A `1.0 → 1.1` migration is added so worlds and compendiums that still carry orphan values get them stripped from disk. The migration walks world items, compendium items, and actor-embedded items, and emits a Foundry `-=key` delete for whichever of `path`, `ancestry`, `power` is physically present on a given talent's `system` data. Talents that already have none of the three keys generate no DB write, so the migration is idempotent and safe to re-run.

**Related Issue**
Closes #314

**How Has This Been Tested?**
- `tsc --noEmit` is clean across `src/` (remaining errors come from `foundry-vtt-types` / `@types/node` and pre-exist on `main`).
- Code paths verified by inspection: no remaining reader of `talent.system.{path,ancestry,power}` or of the derived flags; the `system.ancestry` reference on `ActionItem` and the `actor.ancestry`/`item.isAncestry()` accessors are different concerns and left untouched.
- Migration logic mirrors the structure of `0.3-1.0.ts` (raw sources → per-doc changes object → `updateSource` + `update({ diff: false })`).
- Has not yet been exercised against a live Foundry instance — worth a manual pass on a world with existing talent items before merge.

N/A — no UI change beyond the removal of three orphan input groups on the Talent details tab.

**Checklist:**
- [ ] I have tested my changes on Foundry VTT version: _pending manual verification_


src/system/utils/migration/migrations/1.0-1.1.ts
stanavdb, line 1: Two issues here; First of, a migration isn't really necessary; Fields that are not part of the schema get cleaned during data preparation. Migrations are only needed when the old data is no longer compatible with the new schema.

Secondly (and much less important here), this migration is from `1.0` to `1.1`, but this changes is being introduced in `2.2`.

I'd recommend just dropping the migration entirely.

stanavdb (changes requested):
This looks mostly good to me, aside from the 1 comment. However, I do have a note:
  
> * Has not yet been exercised against a live Foundry instance — worth a manual pass on a world with existing talent items before merge.

In the future please test any changes in Foundry first, before submitting for review.

src/system/utils/migration/migrations/1.0-1.1.ts
mirandaguillaume: I wasn't sure about the migrations, I kind of winged it !

mirandaguillaume:
> This looks mostly good to me, aside from the 1 comment. However, I do have a note:
> 
> > * Has not yet been exercised against a live Foundry instance — worth a manual pass on a world with existing talent items before merge.
> 
> In the future please test any changes in Foundry first, before submitting for review.

That was an artefact of the old PR description, I tested it in a live Foundry afterward, sorry !

stanavdb:
@mirandaguillaume Hey there! Sorry that this has sat around for a bit, but we're updating our contributing guidelines to be clear and explicit about aligning with Brotherwise's no-AI policy. We will not be accepting any PRs created using generative AI tools. This will apply to all PRs going forward (including open ones like this one).  
  
Please confirm whether any generative AI tools were used in the creation of this PR. If none were used, please add the following checklist item to the PR description:

The same applies for your other PR #726. Thanks!

Khunkurisu:
Treating as abandoned due to lack of response to no-AI-use confirmation in over a month.

PR #721, opened 2026-05-18
Feat: Move Erudition Talent Macro

### LINKED ISSUE #718: Move Erudition Talent macro from Stormlight Handbook to System
The macro for the Erudition talent is currently only in the Stormlight Handbook. We know that all future settings are going to have the Scholar path with its Erudition key talent so it makes more sense to have the macro for it be in the system rather than implemented individually for each setting. This would also allow the Erudition talent in the Starter Rules to utilize the macro when the Stormlight Handbook module is not installed

Move the macro from the stormlight handbook module directly into the system

**Version**  
All

**Checklist:**

boxfriend:
**Type**  

- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
This moves the erudition talent macros to the core system from the stormlight handbook

**Related Issue**  
Closes #718 

**How Has This Been Tested?**  
Used each macro in foundry

**Checklist:**  

**Additional context**  
I have a couple of concerns that I'm not sure whether they will affect the review. 
1. exporting the existing FLAGS const from erudition/index.ts, this does end up making it available in foundry as `cosmereRPG.utils.macros.talents.erudition.FLAGS` which is probably not ideal. if necessary I can put the keys directly into the actor.ts file rather than importing FLAGS there
2. I did need to use AnyObject in a couple of places that any was originally being used, but that was also not possible to use in one place that I actually ended up disabling the no-explicit-any linting for a line because it's not clear what object should be there (if i should declare some interface or whatever to be used in any of these places i can do that)

### DIFF
--- a/src/system/documents/actor.ts
+++ b/src/system/documents/actor.ts
@@ -11,6 +11,7 @@ import {
     Size,
     RestType,
     ImmunityType,
+    AttributeGroup,
 } from '@system/types/cosmere';
 import { Talent, TalentTree } from '@system/types/item';
 import {
@@ -56,6 +57,7 @@ import { containsExpertise } from '@system/utils/actor';
 import { SYSTEM_ID } from '@system/constants';
 import { HOOKS } from '@system/constants/hooks';
 import { AnyObject } from '@league-of-foundry-developers/foundry-vtt-types/utils';
+import { FLAGS } from '@system/utils/macros/talents/erudition';
 
 export type CharacterActor = CosmereActor<ActorType.Character>;
 export type AdversaryActor = CosmereActor<ActorType.Adversary>;
@@ -1431,6 +1433,14 @@ declare module '@league-of-foundry-developers/foundry-vtt-types/configuration' {
                 'goals.hide-completed': boolean;
                 [key: `meta.update.mode.${string}`]: string;
                 [key: `mode.${string}`]: string;
+                [FLAGS.SKILLS_COUNT]: number;
+                [FLAGS.SKILLS_GROUPS]: AttributeGroup[];
+                [FLAGS.SKILLS_INCREASE]: number;
+                [FLAGS.SKILLS_SELECTED]: Skill[];
+                [FLAGS.EXPERTISES_COUNT]: number;
+                [FLAGS.EXPERTISES_TYPES]: ExpertiseType[];
+                [FLAGS.EXPERTISES_SELECTED]: Expertise[];
+                [key: `skills.${string}.temporaryRanks`]: number;
             };
         };
 
--- a/src/system/documents/actor.ts
+++ b/src/system/documents/actor.ts
@@ -1436,7 +1436,10 @@ declare module '@league-of-foundry-developers/foundry-vtt-types/configuration' {
 
                 /**
                  * The following keys belong to the Erudition macro
-                 * they were ported over from the Stormlight Handbook as is
+                 * they were ported over from the Stormlight Handbook as is.
+                 * Declaring them in erudition/index.ts was not allowing the Actor
+                 * definition to be merged and a proper way to merge them will need
+                 * to be researched to make this cleaner
                  */
                 [FLAGS.SKILLS_COUNT]: number;
                 [FLAGS.SKILLS_GROUPS]: AttributeGroup[];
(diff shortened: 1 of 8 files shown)


src/system/utils/macros/talents/erudition/dialog.ts
```diff
@@ -0,0 +1,312 @@
...
+        classes: ['dialog', 'erudition'],
+        tag: 'dialog',
+        position: {
+            width: 500,
+        },
+    };
+
+    static PARTS = {
+        form: {
+            template: `modules/cosmere-rpg-stormlight-handbook/templates/dialogs/erudition.hbs`,
```
stanavdb: Reference to the `cosmere-rpg-stormlight-handbook` module

src/system/documents/actor.ts
```diff
@@ -1431,6 +1433,14 @@ declare module '@league-of-foundry-developers/foundry-vtt-types/configuration' {
...
                 [key: `meta.update.mode.${string}`]: string;
                 [key: `mode.${string}`]: string;
+                [FLAGS.SKILLS_COUNT]: number;
+                [FLAGS.SKILLS_GROUPS]: AttributeGroup[];
+                [FLAGS.SKILLS_INCREASE]: number;
+                [FLAGS.SKILLS_SELECTED]: Skill[];
+                [FLAGS.EXPERTISES_COUNT]: number;
+                [FLAGS.EXPERTISES_TYPES]: ExpertiseType[];
+                [FLAGS.EXPERTISES_SELECTED]: Expertise[];
+                [key: `skills.${string}.temporaryRanks`]: number;
```
stanavdb, line 1451: Since these are really only used by the erudition macro I think it'd be better to define these there. Should be able to do:
```ts
declare module '@league-of-foundry-developers/foundry-vtt-types/configuration' {
    interface FlagConfig {
        Actor: {
            'cosmere-rpg': {
                [FLAGS.SKILLS_COUNT]: number;
                [FLAGS.SKILLS_GROUPS]: AttributeGroup[];
                [FLAGS.SKILLS_INCREASE]: number;
                [FLAGS.SKILLS_SELECTED]: Skill[];
                [FLAGS.EXPERTISES_COUNT]: number;
                [FLAGS.EXPERTISES_TYPES]: ExpertiseType[];
                [FLAGS.EXPERTISES_SELECTED]: Expertise[];
                [key: `skills.${string}.temporaryRanks`]: number;
            }
        }
    }
}
```

src/system/documents/actor.ts
```diff
@@ -11,6 +11,7 @@ import {
     Size,
     RestType,
     ImmunityType,
+    AttributeGroup,
```
stanavdb, line 14: Unrelated addition?

stanavdb (changes requested):

src/system/documents/actor.ts
boxfriend: The AttributeGroup type is used in the FlagConfig interface, it's the type associated with the FLAGS.SKILL_GROUPS key

src/system/utils/macros/talents/erudition/dialog.ts
boxfriend: completely forgot about the template, that's fixed now

src/system/documents/actor.ts
boxfriend: This actually was originally included in `erudition/index.ts` however the error I had shown in discord where you showed the FlagConfig in actor.ts persisted, and there was an error about Actor being defined already. The only fix I had found for that was defining those keys in actor.ts because I wasn't able to figure out how to merge them (if that is possible)

src/system/documents/actor.ts
stanavdb: Yep you're right. Looked right over it.

src/system/documents/actor.ts
stanavdb: I think there should be a way to get the declarations to merge nicely, but it's not worth spending time on. Add an explainer comment in actor that these belong to the erudition macro and why they can't be defined there, then it's good as far as I'm concerned.

boxfriend: explanation comment added

stanavdb (approved):

PR #726, opened 2026-05-20
fix(sheets): preserve prose-mirror editor content across re-renders

### LINKED ISSUE #211: Item Description Fields not saving when editing other item fields.
**Describe the bug**  
When editing description and switching to another tab when you switch back description editing screen remains open and is reset to whatever it was before starting editing. So items that had descriptions already go back to its prior description and new items go back to blank. 

**To Reproduce**  

1. Open Any time, new or already created. (I tried several types like equipment/power/loot and it had the same effect.
2. Edit Description.
3. Update any other field in the item IE: title, details, effects, etc

**Expected behavior**  
Description should leave edit mode

**Screenshots**  
https://github.com/user-attachments/assets/3d18664d-b4a2-45b7-a3f7-421ac4762032

**Foundry VTT Version**  
Not mentioned by reporter

0.2.2

**Browser**  
Accessed through browser and only seems to show up when on hosted browser rather than a local foundry instance

**Additional context**  
_Add any other context about the problem here._

**Checklist:**

### LINKED ISSUE #724: Actor biography/appearance/notes editor loses typed content when editing other actor fields
**Describe the bug**

On Character and Adversary sheets, opening the rich-text editor on `biography`, `appearance`, or `notes` and then modifying any other field on the sheet (for example the actor's name in the header) silently discards the typed-but-not-saved content. The editor visually reverts to the previously-saved value, and the typed text is also lost from the actor document — so the work disappears both on-screen and in storage.

This is the actor-side twin of #211 (which only covers the item description editor). Reproduced experimentally on a clean Character actor.

**To Reproduce**

1. Open a Character or Adversary sheet, make sure Edit mode is on.
2. Go to the **Notes** tab.
3. Click the edit (feather) icon next to **Biography** (or Appearance, or Notes).
4. Type some text in the ProseMirror editor — do **not** click the Save button in the editor toolbar.
5. Modify another field on the sheet, e.g. the actor's name in the header. Click elsewhere to blur it.
6. Observe: the editor stays open, its content reverts to the previously-saved value, and the typed text is gone from `actor.system.biography` as well (verified in console: `game.actors.get(id).system.biography`).

**Expected behavior**

The typed text in the editor should survive a re-render triggered by a sibling form field. Either the editor preserves the live content visually across re-renders, or it commits the content before exiting edit mode. In any case, the user's in-flight typing should not be silently discarded.

**Screenshots**
https://github.com/user-attachments/assets/24441c9f-ecf2-4223-b0a0-87290ff67ca5

**Foundry VTT Version**

13 (verified on build 351)

2.1.0

**Browser**

Reproduced in Chromium. The bug is architectural and not browser-specific (unlike #211 which the original reporter only saw on hosted Foundry; this one shows up everywhere).

**Additional context**

Root cause is the same architectural pattern as #211: `BaseActorSheet` stores a snapshot of the field's HTML in `proseFieldHtml` at the moment the user clicks the edit icon. This snapshot is never refreshed, and each re-render builds the new `<prose-mirror>` element from it — so any in-flight typing is overwritten visually on the next render.

Additionally, `BaseActorSheet.onFormEvent` performs single-field updates (`change = {[event.target.name]: ...}`), so the live `pm.value` is not sent to `actor.update` when another field changes. The typed content is therefore lost from persistence too. (On the item side the loss is purely visual because `BaseItemSheet` pushes the whole `formData`, which does include the live `pm.value`. The actor's narrower update widens the data loss.)

The fix being proposed for #211 — capturing the live `<prose-mirror>` value in `_prepareContext` before the re-render replaces the DOM — applies almost verbatim here on `BaseActorSheet._prepareContext`, targeting `prose-mirror[name="${this.proseFieldName}"]`. Happy to open a follow-up PR once #211 lands.

**Checklist:**

mirandaguillaume:
**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
Fixes a stale-snapshot bug in `BaseItemSheet` where in-flight prose-mirror content was lost on re-render when a sibling field change triggered `submitOnChange` (#211). The same root cause was applying to `BaseActorSheet` for the `biography`, `appearance`, and `notes` fields (#724) — and there the data loss also reached persistence, because `onFormEvent` performs a narrow single-field update rather than pushing the full `formData`, so the stale snapshot was being committed back to the actor document.

The fix syncs the in-memory snapshot (`proseDescHtml` on items / `proseFieldHtml` on actors) from the live DOM element at the top of `_prepareContext`, before the re-render replaces the editor. A `querySelector` on `prose-mirror[name="<fieldName>"]` targets only the currently active editor so unrelated fields are not touched.

Files changed:
- `src/system/applications/item/base.ts` — sync `proseDescHtml` in `_prepareContext`; tighten the `onFormEvent` element-type guard; fix the `item-description-tab.hbs` condition to `(and isUpdatingDescription tab.active)`
- `src/system/applications/actor/base.ts` — sync `proseFieldHtml` in `_prepareContext` (same pattern)

**Related Issue**  
Closes #211  
Closes #724

**How Has This Been Tested?**  
Manual reproduction of both bugs on Foundry VTT v13.351, before and after the fix:

1. Item sheet: open any item, click into a description field, type text without saving, then change the item name in another input. Before the fix the typed content disappears on re-render; after the fix it survives.
2. Actor sheet: open a Character or Adversary, click into Biography / Appearance / Notes, type text without saving, then change the actor name in another input. Before the fix the typed content disappears *and* the empty value is persisted to `actor.system.<field>` (verifiable in the browser console). After the fix both the visible editor content and the stored value are preserved.
3. Saving via the prose-mirror toolbar button still commits content normally — no regression to the happy path.
4. Verified the `querySelector` only matches the active editor: editing one prose-mirror field does not touch the snapshot of another.

CI/local checks:
- `tsc --noEmit` — passes
- `npm run build` — passes (baseline unchanged)
- `npm run lint` — passes (baseline unchanged)

N/A — the bug is invisible by definition (content disappears). Reproduction is described in the testing steps above.

**Checklist:**  

**Additional context**  
The fix is purely defensive — it reads the live DOM before the framework discards it — and has no effect when no prose-mirror field is currently being edited. Two related issues (#211 and #724) share the same root cause, hence one PR closing both.


stanavdb:
Please update this to conform to our PR template.

mirandaguillaume:
> Please update this to conform to our PR template.

Done !

Khunkurisu (approved):

Khunkurisu (changes requested):
I had initially approved this cause it seemed to work correctly. And it does, but it has a minor oversight. If you try to save your work after having performed an update, despite the text field showing the correct values, it will not save properly. Editing the text that was preserved after the update does correct this, and it will save afterwards.

mirandaguillaume:
Good catch @Khunkurisu, thanks — you were right, the field showed the correct text but it wasn't actually being written to the document.

Root cause: the original approach preserved the in-progress text by re-feeding it as the rebuilt `<prose-mirror>`'s `value`. That makes the new element's committed value (`_value`) equal to its content, so Foundry's internal `#save()` short-circuits (it only fires `change` when `value !== _value`) and the `change` event that `submitOnChange` relies on to persist the field never fires. Re-editing dirtied the editor again, which is exactly why typing then saving "fixed" it.

New approach: instead of capturing and replaying the editor across the render, I now suppress the re-render while a prose-mirror editor holds unsaved content (`_canRender()` + `isDirty()`), the same way core's `JournalEntryPageProseMirrorSheet` does. The editor is no longer destroyed mid-edit, so the original content-reset issue stays fixed and a plain save persists correctly. I also dropped the now-redundant `_prepareContext` capture blocks on both the item and actor sheets.

Verified the edit → external update (re-render) → save flow now persists on a v13 world. Ready for another look whenever you have a moment.

Khunkurisu:
Treating as abandoned due to lack of response to no-AI-use confirmation in over a month.

PR #761, opened 2026-06-02
Feat: Create Effects Container Document

### LINKED ISSUE #757: Effect Container Document That Functions as Parent In Effects List
This has been discussed multiple times, and v14 will partially make this redundant. Which makes a solid stop-gap *more* valuable not less, as it'll allow us to prototype features related to having effects be a proper document and get users used to it, prior to the v14-compatible-release.

Just a simple item type that is a container for effects, similar to the 3.0.0 ability for actions to embed effects in.

This needs to be done sooner or later, because of a multitude of reasons and it is by far one of the easiest things to adapt & edit of all our current "I want this for ease of use" features. There is no alternative right now, just getting this in quickly and adapting it as needs arise is absolutely 100% the right move.

**Version**  
2.1.0

**Additional context**  
Again, v14 will partially make this obsolete but I will be maintaining it with that in mind regardless, and waiting until we find time to port to v14 makes no sense anymore.

**Checklist:**

Khunkurisu:
**Type**  

- [ ] Bug fix
- [x] Feature
- [x] Refactor

**Description**  
Adds a new item type "Effects Container" which has description fields, an id, and can hold events as well as the ability to contain effects. Also adjusts the effects tab to make use of the item list functionality added in 2.1.0, thus refactoring a lot of the logic for how that tab is displayed. Effects will now show in their appropriate list either as-normal or indented beneath their parent item, depending on if that item has more than one effect in that section. Parent items can be expanded, and a context menu option has been added to child effects to view their parent's sheet. Empty effects containers will appear in the Inactive Effects section when added to an actor sheet so that they can be edited by the user.

**Related Issue**  
Closes  #757

**How Has This Been Tested?**  
Add effects containers to an actor, note that if they are empty they appear in inactive effects. Add effects to the containers and any other items on the sheet, in varied count. Open context menus and click the list item for containers to ensure all actions function as expected. Examine the styling to ensure no regressions were introduced.

imageimage

**Checklist:**  

**Additional context**  
Note: The files changed looks more daunting than it really is. I added a generic type that expresses schema return objects more effectively, as well as a TEMPLATES constant for the system templates folder. These are not necessary for the functionality of the PR, they were just random improvements I chose to make while in the process of competing the task.


Doc-Sun (approved):
I submit my own review from a data entry/user end. This was super smooth and works well. Its backwards compatible with older some adversaries that are complicated and have multiple effects off of single items. It just works exactly as described and smoothly.

src/system/applications/actor/components/effects-list.ts
```diff
@@ -66,39 +72,90 @@ export class ActorEffectsListComponent extends HandlebarsApplicationComponent<
...
+                // handle item state & data
+                this.itemState[item.id!] = { expanded: false };
+                this.itemData[item.id!] = {
+                    ...(item.hasDescription() && item.system.description?.value
+                        ? {
+                              descriptionHTML: await TextEditor.enrichHTML(
+                                  item.system.description.value,
+                                  {
+                                      relativeTo: (item as CosmereItem).system
+                                          .parent as foundry.abstract.Document.Any,
```
zithith, line 137: could this be cast as `CosmereDocument` or maybe more specifically `CosmereItem` as we shouldn't be getting Actors in this list?

src/system/data/fields/string-field.ts
zithith, line 1: Was this meant to be left commented out? Can we re-instate or just remove it?

zithith (approved):

src/system/applications/actor/components/effects-list.ts
Khunkurisu: Shouldn’t be a problem, no. I’d just copied the basic logic out of the parent list item class without investigating how the description was being grabbed so left it untouched.

Khunkurisu: That said, this will need changed in the 3.0.0 port regardless due to other item list necessities there so I’m willing to leave the unnecessary cast there for now since it’s not affecting anything but transparency.

src/system/data/fields/string-field.ts
Khunkurisu: This has been commented out since the v13 changes so I can only imagine we’d need a host of changes to reinstate it, so I just deleted the file as it’s not done anything in a long time.

PR #734, opened 2026-05-25
feat(items): integrate resource consumption with embedded actions

stanavdb:
**Type**  
- [ ] Bug fix
- [x] Feature
- [x] Refactor
- [ ] Other (please describe):

**Description**  
This PR refactors the activation consumption system to add support for item resource and to integrate with the embedded actions approach.

It restructures the underlying consumption model, updates related action/use flow, and updates the UI/UX around configuring resource consumption and item resources. It also adds a reusable document-matching utility similar to what's used in the event system; Currently only used for resource consumption but should eventually replace document matching in the event system as well.

**Related Issue**  
Partially addresses #358, Closes #735, Closes #736

**How Has This Been Tested?**  
_Item resource configuration_
1. Create a weapon item
2. Navigate to the "Details" tab
3. Click the "Add Resource" select input and pick "Uses"
4. Ensure the "Uses" resource is added with value `1`, max `1`, and recharge `none`
5. Ensure primary resource has switched to "Uses"
6. Change max to `3`
7. Change value to `2`
8. Change recharge to `Per Scene`
9. Reload the client and ensure all changes have persisted
10. Change value to `5` -> ensure max is increased to `5`
11. Change max to `2` -> ensure value is lowered to `2`
12. Click the "Primary Resource" select and ensure "Uses" is the only option
13. Click the "Add Resource" select input -> Ensure "Uses" is not in the list of options
14. Pick "Charges" 
15. Ensure the "Charges" resource is added with value `1`, max `1`, and recharge `none`
16. Ensure the "Add Resource" select input is no longer visible
17. Change value for the "Charges" resource to `2`
18. Change the primary resource to "Charges"
19. Reload the client and ensure the change has persisted
20. Change the primary resource to "Uses"
21. Remove the "Uses" resource 
22. Ensure the primary resource has switched to "Charges"
23. Ensure the "Add Resource" select is visible again and provides the "Uses" option

_Resource consumption_
1. Create weapon item
2. Create character actor
3. Add weapon to actor
4. Edit the weapon
1. Open the "Actions" tab
5. Add a new action
6. Navigate to the "Details" tab
7. Set "Activation Type" to `Utility`
8. Set "Activation Cost" to `Reaction`
9. Add a new resource consumption entry
10. Close edit resource consumption dialog
11. Ensure resource consumption entry is added and reads: 1 **Focus** from _<reference to actor>_
12. Set actor focus to `2`
13. Set actor health to `10`
14. Use created action from actor sheet
[...]

PR #766, opened 2026-06-09
Feat/Event triggers for combat Start/End & Round Start/End

### LINKED ISSUE #651: Add Combat Encounter Started/Ended Event Trigger

dtfe:
**Type**  

- [ ] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  
This PR implements event triggers for when Combat starts or ends, as well as when a round ends or starts.
Uses the built-in Foundry hooks combatStart & deleteCombat for Combat Start and End triggers
Uses custom hooks for Round Start/End which include round number data for the previous round and the new round.

**Related Issue**  
Closes #651 and #652 

**How Has This Been Tested?**  
Combat Start/End
1. Create new character
2. Navigate to "Equipment" section
3. Create new equipment
4. Navigate to Events
5. Create new event
6. Set trigger to "Combat Started"
7. Set handler to "Set Attribute", awareness, 5
8. Press update
9. Create new event
10. Set trigger to "Combat Ended"
11. Set handler to "Set Attribute", presence, 5
12. Press update
13. Create encounter
14. Add new character to encounter
15. Press begin encounter
16. Awareness should be 5
17. Press end encounter
18. Presence should be 5

Round Start
1. Create new character
2. Navigate to "Equipment" section
3. Create new equipment
4. Navigate to Events
5. Create new event
6. Set trigger to "Combat Round Started"
7. Set handler to "Set Attribute", strength, 5
8. Press update
9. Create encounter
10. Add new character to encounter
11. Press begin encounter (Round 1)
12. Press next Round (Round 2)
13. See strength having increased to 5. 
14. Open console
15. Paste `Hooks.on("cosmere-rpg.combatRoundStart", (...args) => {console.log("combatRoundStart fired", args);});`
16. Press next round (Round 3) 
17. Console should contain hook information. Expand options. "(newRound: 3, previousRound: 2)"

Round End (Continue from previous):
18. Create new event
19. Set trigger to Combat Round Started
20. Set handler to Set Attribute, speed, 5
21. Press update
22. Open console
23. Paste `Hooks.on("cosmere-rpg.combatRoundEnd", (...args) => {console.log("combatRoundEnd fired", args);});`
24. Press next round (round 4)
25. Agility should be set to 5
26. console should contain hook information. Expand options "(newRound: 4, previousRound 3)"
27. RoundEnd hook should have triggered first and RoundStart triggered after.

Round Previous (Continue from previous):
28. Create new event
29. Set trigger to Combat Round Previous
30. Set handler to Modify Skill Rank, Agility, -1
31. Press update
32. Go to skills section
33. Set agility to 5
34. Press previous round
35. Agility should be reduced to 4.

image
image

**Checklist:**  
- [X] I have tested my changes on Foundry VTT version: 13 Build 351.

### DIFF
--- a/src/system/constants/hooks.ts
+++ b/src/system/constants/hooks.ts
@@ -14,6 +14,13 @@ export const HOOKS = {
     PRE_REST: `${SYSTEM_ID}.preRest`,
     REST: `${SYSTEM_ID}.rest`,
 
+    /* ----- Encounter Hooks ----- */
+    /* -- Combat Hooks -- */
+    COMBAT_START: `combatStart`,
+    COMBAT_END: `deleteCombat`,
+    COMBAT_ROUND_START: `${SYSTEM_ID}.combatRoundStart`,
+    COMBAT_ROUND_END: `${SYSTEM_ID}.combatRoundEnd`,
+
     /* ----- Item Hooks ----- */
     /* -- Item Activation/Usage Hooks -- */
     USE_ITEM: `${SYSTEM_ID}.useItem`,
@@ -58,7 +65,8 @@ export const HOOKS = {
     /* ----- Roll Hooks ----- */
     PRE_ROLL: (context: string) =>
         `${SYSTEM_ID}.pre${context.toLowerCase().capitalize()}Roll` as const,
-    ROLL: (context: string) => `${SYSTEM_ID}.${context.toLowerCase()}Roll` as const,
+    ROLL: (context: string) =>
+        `${SYSTEM_ID}.${context.toLowerCase()}Roll` as const,
     PRE_SKILL_ROLL: `${SYSTEM_ID}.preSkillRoll`,
     SKILL_ROLL: `${SYSTEM_ID}.skillRoll`,
     PRE_ITEM_ROLL: `${SYSTEM_ID}.preItemRoll`,
--- a/src/system/constants/hooks.ts
+++ b/src/system/constants/hooks.ts
@@ -16,8 +16,6 @@ export const HOOKS = {
 
     /* ----- Encounter Hooks ----- */
     /* -- Combat Hooks -- */
-    COMBAT_START: `combatStart`,
-    COMBAT_END: `deleteCombat`,
     COMBAT_ROUND_START: `${SYSTEM_ID}.combatRoundStart`,
     COMBAT_ROUND_END: `${SYSTEM_ID}.combatRoundEnd`,
 
--- a/src/system/hooks/item-event-system/events.ts
+++ b/src/system/hooks/item-event-system/events.ts
@@ -106,6 +106,36 @@ const EVENTS: EventDefinition[] = [
             duration === RestType.Long,
         transform: (actor: CosmereActor) => ({ document: actor }),
     },
+
+    // Combat events
+    {
+        type: 'combat-encounter-start',
+        hook: HOOKS.COMBAT_START,
+        transform: (combat: Combat) => ({ document: combat }),
+    },
+    {
+        type: 'combat-encounter-end',
+        hook: HOOKS.COMBAT_END,
+        transform: (combat: Combat) => ({ document: combat }),
+    },
+    {
+        type: 'combat-round-start',
+        hook: HOOKS.COMBAT_ROUND_START,
+        transform: (combat: Combat, updateData: { round: number }) => ({
+            document: combat,
+            previousRound: combat.round,
+            newRound: updateData.round,
+        }),
+    },
+    {
+        type: 'combat-round-end',
+        hook: HOOKS.COMBAT_ROUND_END,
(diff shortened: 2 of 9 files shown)


src/system/constants/hooks.ts
```diff
@@ -14,6 +14,13 @@ export const HOOKS = {
     PRE_REST: `${SYSTEM_ID}.preRest`,
     REST: `${SYSTEM_ID}.rest`,
 
+    /* ----- Encounter Hooks ----- */
+    /* -- Combat Hooks -- */
+    COMBAT_START: `combatStart`,
+    COMBAT_END: `deleteCombat`,
```
stanavdb: This file is used to define system hooks, we shouldn't be re-defining native foundry hooks here.  
If the problem is that the events hook field is typed as `string` then we should solve that instead.

src/system/hooks/item-event-system/events.ts
```diff
@@ -106,6 +106,36 @@ const EVENTS: EventDefinition[] = [
...
+        transform: (combat: Combat) => ({ document: combat }),
+    },
+    {
+        type: 'combat-round-start',
+        hook: HOOKS.COMBAT_ROUND_START,
+        transform: (combat: Combat, updateData: { round: number }) => ({
+            document: combat,
+            previousRound: combat.round,
+            newRound: updateData.round,
+        }),
```
stanavdb: Seemingly two issues here: 
1. The hook transform handler seems to have the wrong type. The hook is invoked as `Hooks.callAll(HOOKS.COMBAT_ROUND_START, this, previousRound, newRound);`, which doesn't have the `updateData`. I don't think this is working as expected, please verify.
2. The transform function serves to take a system defined hook signature (which can have arbitrary params) and transform it to be standardized (document, options, userId). This means the only used output fields are `document`, `options`, `userId`. Anything else is simply ignored. `previousRound` and `newRound` do nothing here.

Before you re-request review: Please verify these are fully working as expected with a test script included in the PR. See #734 for test script reference.

src/system/hooks/item-event-system/events.ts
```diff
@@ -106,6 +106,36 @@ const EVENTS: EventDefinition[] = [
...
+    },
+    {
+        type: 'combat-round-end',
+        hook: HOOKS.COMBAT_ROUND_END,
+        transform: (combat: Combat, updateData: { round: number }) => ({
+            document: combat,
+            previousRound: updateData.round,
+            newRound: combat.round,
+        }),
```
stanavdb: Same as above

src/system/types/hooks/combat.ts
```diff
@@ -0,0 +1,9 @@
+import { HOOKS } from '@system/constants/hooks';
+
+declare module '@league-of-foundry-developers/foundry-vtt-types/configuration' {
+    namespace Hooks {
+        interface HookConfig {
+            [HOOKS.COMBAT_START]: [combat: Combat];
```
stanavdb: If `combatStart` is not an existing foundry hook, it lacks invocation (this PR doesn't add it).  
If `combatStart` IS an existing foundry hook, it shouldn't need to be declared here. `foundry-vtt-types` should be declaring it.  
If `combatStart` IS an existing foundry hook, but `foundry-vtt-types` doesn't declare it, please add an explanatory comment so it's clear why this declaration is necessary.

stanavdb (changes requested):

dtfe:
I think that covers all your requested changes

stanavdb (commented):
Something about these changes is breaking the build, causing Foundry to fail to load the world/system.  Only occurring on this branch for me. Is it working for you?

Image
Image

dtfe:
> Something about these changes is breaking the build, causing Foundry to fail to load the world/system. Only occurring on this branch for me. Is it working for you?
> 
> Image Image

Yeah it was working on my end when I changed it but then when I retried it now it seems to break the build. A line I forgot to remove after deleting the unnecessary src/system/types/hooks/combat.ts file. Thanks for catching that. Regarding the first error, that seems to exist on branch Release 2.2.0, so it must be related to something else.

stanavdb (changes requested):
Build works now. Some things I noticed from testing:
- The combat round start hook doesn't trigger on the first round (when combat itself is started). I would expect it to trigger there too.
- The combat round start and combat round end hooks don't work with going back to previous rounds, only when advancing to the next round.
- The combat round start and combat round end hooks are functionally identically; They fire at the exact same time and pass the exact same information. Maybe a combat round change hook would be better? Alternatively could go for next/previous round hooks, though a round change hook would contain all the relevant info already.

Khunkurisu (approved):

Khunkurisu:
Meant to post as part of review... Since the changes requested by Stan seemed resolved, and this is otherwise functional I'm approving it. Further changes can come later if somehow we missed something.

stanavdb (approved):

PR #768, opened 2026-06-10
Feat/Basic language expertise

### LINKED ISSUE #362: Expose languages array to character sheet
Cultures in the source material grant access to reading and writing languages and there are other modules that add features related to the languages that a character may be able to read or write in. Currently there is no built in way to add languages to the existing languages array on a character.

A checkbox like the expertises exposed to the character sheet or even selecting a culture could add languages to the array for use with these other modules (such as Polyglot)

Current workarounds include adding items to a character sheet with the language name, but this is inelegant and just clutters the sheet more. It is also possible to use effects to add languages to the array and this could even be the solution built into the cultures in the compendium. These options do work so the feature isn't necessary, but of course workarounds aren't that great and a native feature that integrates with the system better would be preferable.

**Version**  
0.3.1

**Additional context**  
_Add any other context or screenshots about the feature request here._

**Checklist:**
- [✅] I have searched the existing issues to ensure this feature has not already been requested.

dtfe:
**Type**  

- [ ] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  
Only includes a language category for the expertise. Does not update cultures with 

**Related Issue**  
Closes #362 

**How Has This Been Tested?**  
1. Create new world
2. Create new actor
3. Open actor
4. Edit expertises
5. See language expertise
6. Add any language of your choice and see it appear in the expertise section
7. Go to compendium
8. Go to cultures
9. Drag and drop each one and see that it gives you the respective language
10. Remove each one and see it takes away the respective language

image
image

**Checklist:**  
- [X] I have tested my changes on Foundry VTT version: V13 Build 351.

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -20,6 +20,7 @@
                     "name": "Expertise",
                     "Armor": "Armor",
                     "Cultural": "Cultural",
+                    "Language": "Language",
                     "Specialist": "Specialist",
                     "Utility": "Utility",
                     "Weapon": "Weapon"
--- a/src/packs/cultures/alethi.json
+++ b/src/packs/cultures/alethi.json
@@ -22,6 +22,12 @@
                             "type": "cultural",
                             "label": "Alethi",
                             "locked": false
+                        },
+                        "language:lang_alethi": {
+                            "id": "lang_alethi",
+                            "type": "language",
+                            "label": "Alethi",
+                            "locked": false
                         }
                     }
                 }
@@ -36,6 +42,12 @@
                         "cultural:alethi": {
                             "id": "alethi",
                             "type": "cultural",
+                            "label": "Alethi",
+                            "locked": false
+                        },
+                        "language:lang_alethi": {
+                            "id": "lang_alethi",
+                            "type": "language",
                             "label": "Alethi",
                             "locked": false
                         }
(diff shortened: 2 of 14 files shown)


stanavdb (changes requested):
If we add a new expertise category, we also need to add the related `CONFIG` so standard values (alethi, azish, etc.) can be configured and the appropriate API function to go with it. Additionally if we add a language expertise category it would need to integrate with cultures automatically so that the appropriate languages are granted.

I think the core issue of #362 is that the actor data model provides a `languages` field ([link](https://github.com/the-metalworks/cosmere-rpg/blob/3f20b7877cbc25adb7195d97f8458afd7fba4dbd/src/system/data/actor/common.ts#L155)) but that field isn't populated by anything as of now. The easiest solution there is just to derived the value of the languages field from the cultural expertises.

dtfe:
I decided to stick to expertise. As the book talks about language underneath expertise, I think it makes the most sense to put it in expertise. If something needs to get the actor's languages it should be retrievable within expertises.

I think(?) I added the events needed to grant the language expertises when importing the different cultures. It at least worked for me when I rebuilt and packed it.
I'll also do so for the stormlight handbook and PR that as well. 

From my understanding of the id system here they needed unique IDs. I did not test if that was the case, but since I read priority is used in cases where ids are the same, I chose to use "lang_(language name)" just incase it would clash with cultures if they shared the same ID.

dtfe:
The team did not agree with the changes so I decided to close this and leave it be for now.

PR #780, opened 2026-06-19
Fix/Context menu embedded actions fix #745

### LINKED ISSUE #745: 'Remove' Button in Context Menu for Embedded Actions Does Nothing
**Describe the bug**  
In the actor's actions tab, actions have a context menu with a remove button. If the action is embedded on another item, the remove button is present but does nothing. This is because it is trying to find the action by its raw id in the actor's collection, not accounting for a potential parent.

**To Reproduce**  
1. Add an item which can contain actions to an actor sheet
2. Add an action to that item
3. Go to the actor's action tab and click the context menu for the action you just made
4. Click remove...
5. Nothing happen

**Expected behavior**  
It should either remove the action from the parent or not display.

**Screenshots**  
Image
Image

**Foundry VTT Version**  
13.351

|**Name**|**Version**|
|---|---|
|**System**|3.0.0|

**Browser**  
Local foundry install

**Additional context**  
N/A

**Checklist:**

dtfe:
**Type**  

- [X] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
adds "Item" as an embedded document name for Items to keep TS from complaining and prevent building
Modifies the action-list script to check if the item has a parent and if the parent is an Item. If it is an Item it will delete from the parent's embeddedCollection instead of the actor's. Otherwise it will use the normal behavior.

**Related Issue**  
Closes #745  

**How Has This Been Tested?**  
1. Create new actor
2. Give actor Singer ancestry
3. Go to Talents section
4. Edit Change Form
5. Go to actions section of dialogue
6. Add new action
7. Exit dialogue.
8. Navigate to Actions section of actor
9. Open context menu on the new action under Change Form
10. Press Remove
11. See that it has been removed from both Change Form's actions and the actor's action list.

Before pressing delete on context menu in action's section
image

After
image

**Checklist:**  
- [X] I have tested my changes on Foundry VTT version: v13 build 531.

**Additional context**  

### DIFF
--- a/src/declarations/foundry/client/documents/item.d.ts
+++ b/src/declarations/foundry/client/documents/item.d.ts
@@ -1,6 +1,3 @@
 declare namespace Item {
     type Parent = Actor.Implementation | Item.Implementation | null;
-    namespace Embedded {
-        type Name = 'ActiveEffect' | 'Item';
-    }
 }
--- a/src/declarations/foundry/client/documents/item.d.ts
+++ b/src/declarations/foundry/client/documents/item.d.ts
@@ -1,3 +1,6 @@
 declare namespace Item {
     type Parent = Actor.Implementation | Item.Implementation | null;
+    namespace Embedded {
+        type Name = 'ActiveEffect' | 'Item';
+    }
 }
--- a/src/system/applications/actor/components/actions-list.ts
+++ b/src/system/applications/actor/components/actions-list.ts
@@ -489,11 +489,24 @@ export class ActorActionsListComponent extends ActorItemListComponent {
                                 name: 'GENERIC.Button.Remove',
                                 icon: 'fa-solid fa-trash',
                                 callback: () => {
-                                    // Remove the item
-                                    void this.application.actor.deleteEmbeddedDocuments(
-                                        'Item',
-                                        [item.id!],
-                                    );
+                                    // Check if the Item's parent is not the actor
+                                    const parent = item.parent;
+                                    if (
+                                        parent &&
+                                        parent.documentName === 'Item'
+                                    ) {
+                                        // Remove the item from parent Item
+                                        void parent.deleteEmbeddedDocuments(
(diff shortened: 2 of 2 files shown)


src/system/applications/actor/components/actions-list.ts
```diff
@@ -489,11 +489,24 @@ export class ActorActionsListComponent extends ActorItemListComponent {
...
+                                            'Item',
+                                            [item.id!],
+                                        );
+                                    } else {
+                                        // If Item's parent is an Actor, remove the item from the Actor
+                                        void this.application.actor.deleteEmbeddedDocuments(
+                                            'Item',
+                                            [item.id!],
+                                        );
+                                    }
```
stanavdb: This whole thing can be simplified to:

```suggestion
                                    void item.delete();
```

src/declarations/foundry/client/documents/item.d.ts
```diff
@@ -1,3 +1,6 @@
 declare namespace Item {
     type Parent = Actor.Implementation | Item.Implementation | null;
+    namespace Embedded {
+        type Name = 'ActiveEffect' | 'Item';
+    }
```
stanavdb: Unrelated change

stanavdb (changes requested):

src/system/applications/actor/components/actions-list.ts
dtfe: I actually considered this, but I figured it wasn't done this way for one reason or another.

src/declarations/foundry/client/documents/item.d.ts
dtfe: I'll remove it. It was required to build with the original solution as TS was complaining about Item not having "Item" as an embedded document name. Only "ActiveEffect".

stanavdb (approved):

PR #791, opened 2026-06-22
fix: Adds a check to ensure changes.system contains strike

### LINKED ISSUE #790: Cannot create new events in weapon items
**Describe the bug**  
Cannot create new events on weapons

**To Reproduce**  
1. Create a new weapon item
2. Go to events tab
3. Press + to create a new event
4. Error in console and no event is created

**Expected behavior**  
Upon trying to create a new event, a new event should be created

**Screenshots**  
Image

**Foundry VTT Version**  
v13 build 351

|**Name**|**Version**|
|---|---|
|**System**|3.0.0|

**Browser**  
Foundry Client

**Checklist:**
- [X] I have searched the existing issues to ensure this bug has not already been reported

dtfe:
**Type**  

- [X] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
Aims to fix the bug which prevents you from creating events on weapon items

**Related Issue**  
Closes #790 

**How Has This Been Tested?**  
1. Create new weapon item
2. Go to events
3. Press + to create a new event
4. Event is created and no error.

image

**Checklist:**  

### DIFF
--- a/src/system/documents/item.ts
+++ b/src/system/documents/item.ts
@@ -569,7 +569,8 @@ export class CosmereItem<
                 (changes.system?.strike?.skillLocked ||
                     this.system.strike.skillLocked) &&
                 weaponType !== WeaponType.Special &&
-                !!changes.system
+                !!changes.system &&
+                changes.system?.strike
             ) {
                 const strike = foundry.utils.mergeObject(
                     changes.system.strike,
--- a/src/system/documents/item.ts
+++ b/src/system/documents/item.ts
@@ -566,11 +566,10 @@ export class CosmereItem<
             const changes = changed as Partial<WeaponItem>;
             const weaponType = changes.system?.type ?? this.system.type;
             if (
-                (changes.system?.strike?.skillLocked ||
-                    this.system.strike.skillLocked) &&
+                !!changes.system?.strike &&
                 weaponType !== WeaponType.Special &&
-                !!changes.system &&
-                changes.system?.strike
+                (changes.system.strike.skillLocked ||
+                    this.system.strike.skillLocked)
             ) {
                 const strike = foundry.utils.mergeObject(
                     changes.system.strike,


src/system/documents/item.ts
```diff
@@ -569,7 +569,8 @@ export class CosmereItem<
                 (changes.system?.strike?.skillLocked ||
                     this.system.strike.skillLocked) &&
                 weaponType !== WeaponType.Special &&
-                !!changes.system
+                !!changes.system &&
+                changes.system?.strike
```
Khunkurisu: Should probably prefix with `!!` to basically double-invert it, that way it always provides a boolean value to avoid "falsy" values that aren't explicitly typed. It shouldn't be a problem here, but it's common practice in JS/TS because of how JS handles values like this.

Khunkurisu: Note that boolean coercion isn't *necessary* in if statements, so there's nothing about this that is a "must do it this way to pass review".

NoahJGersh: ~~The `&&` is already going to coerce the values to a boolean in this case.~~ Wrote this bit before seeing your follow-up lmao

This is super nitpicky, but I would argue though that because `changes.system` is already a necessary field, it would be good to have it as the first condition. Then we don't need any additional optional chaining either because their nonexistence would short-circuit the condition. And then still keeping the `!!` to make sure the boolean behavior is preserved (rather than truthy coalescing)

I.e.
```ts
!!changes.system?.strike && // short circuits if either changes.system or changes.system.strike are undefined
weaponType !== WeaponType.Special &&
(changes.system.strike.skillLocked || this.system.strike.skillLocked) // no need for optional chaining because changes.system.strike is guaranteed to exist
```

It's a bit easier to read that way, and eliminates some of the uncertainty.

dtfe: nitpick is important for clarity so I'll implement the suggestion. I do agree with you fully

stanavdb (approved):

Issue #358, opened 2025-05-12
Separate Out Action Functionality In Item Data Model


Khunkurisu:
**Is your feature request related to a problem? Please describe.**  
Currently, there can only be a single action + result associated with any given item. There's many situations where such items can provide multiple actions.

**Describe the solution you'd like**  
Instead of including the activation cost, skill test, damage roll, etc for items as part of their data model, the "Actions" item should handle all of that. Then, we can link actions (multiple, singular, or none) to other items. This will result in a couple benefits:
- Items without actions will no longer bloat the Actions tab on the character sheet
- Simplifies multi-resource-cost actions
- Reduces the size of the Details tab on most items, as it would display a link to each action item, rather than the full action details

**Describe alternatives you've considered**  
Our current handling mostly works, it's just a bit janky and under-engineered.

**Version**  
0.3.1

**Additional context**  
N/A

**Checklist:**
- [x] I have searched the existing issues to ensure this feature has not already been requested.

stanavdb:
As an alternative we can also look to split the concept of actions and activations entirely in a similar manner to the way 5e handles its activities. Might be slightly cleaner (as we're not creating a ton of fairly specific Items in the compendium) while still achieving the same result.

Khunkurisu:
> As an alternative we can also look to split the concept of actions and activations entirely in a similar manner to the way 5e handles its activities. Might be slightly cleaner (as we're not creating a ton of fairly specific Items in the compendium) while still achieving the same result.

This was discussed as an example, yeah. There’s likely a compromise of both options that works even better than either alone would, tbh. Depends on how we wanna design it.

Issue #778, opened 2026-06-17
Integrate item resources into "Cost" section of character sheet Actions tab


posadist-revolution:
**Is your feature request related to a problem? Please describe.**  
After #734 merges, we can create actions which use item resources, but those item resources currently do not have a way to display on the "Cost" section

**Describe the solution you'd like**  
Create and display some listing of the costs

**Describe alternatives you've considered**  
Not having this display

**Version**  
3.0.0

**Additional context**  
Long-term I think we'd like "Item Resources" to be able to create a "choose resource target" prompt to choose between items if there are multiple matches with charges, so our UI shouldn't assume there's only 1 item we'll be pulling charges from

**Checklist:**
- [x] I have searched the existing issues to ensure this feature has not already been requested.

PR #798, opened 2026-06-28
feat(migrations): add 2.2 to 3.0 migration for embedded actions

### LINKED ISSUE #738: Add Migration Logic for Actions on Existing Items
As title says, migration will be needed so that all action-related content doesn't break and require data entry be done from scratch. Some things simply won't transfer over perfectly, but the primary goal is to not break anyone's world and to limit extra effort for users and developers alike to transition to the new setup.

stanavdb:
**Type**  
- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
Adds the necessary migration for embedded actions. This migration takes any items that have an activation configured in 2.2 and creates an embedded action with the same activation config.

**Related Issue**  
Partially addresses #358, closes #738

**How Has This Been Tested?**  
1. Check out main branch and build 
2. Create world
3. Create at least 1 actor with ancestry, culture, path, talents, and items (at least 1 weapon)
4. Create at least 1 weapon
5. Close Foundry
6. Check out this PR and build
7. In Foundry load world (create backup for repeats)
8. Ensure no unexpected errors show up in console
9. Ensure all documents are valid
10. Open actor, ensure all actions show as expected

**Checklist:**  


dtfe (changes requested):
## Testing Setup
1. Use stonewalkers with release 2.1.0
2. Create new character
3. Give new character human, Alethi, Warrior
4. Give warrior talents up to Feinting strike
5. Give items 
- Anesthetic (5 doses) and use it 3 times
- Antiseptic (potent, 5 doses) (Do not use) 
- Antiseptic (weak, 5 doses) and use it up until empty
- Drainer (Use once)
- Poison (Do not use)
- Shardplate and use once
6. Get guard from stonewalkers, adversaries and drag token to scene
7. Update to 2.2.0
8. Migrate stonewalkers world
9. See everything is the same
10. Update to 3.3.0 (this branch)
11. Migrate stonewalkers world
12. See below:

## Adversaries
Before Migration
image
<br>
After Migration
image
Debilitate and Work as one both missing resource consumption of focus. Might not be entirely relevant for this PR?

## Uses
Before Migration 
Image
<br>
After Migration  
Image
All items that consume a resource like uses/charges end up with charges when migration ends. 

## Charges
Before Migration
image  
<br>
After Migration
Image
Drainer uses charges, not uses. Yet the migration assumes uses for resource consumption, but charges for resources
image
Same goes for shardplate

## Character Resources
Before Migration
image
<br>
After Migration
Image  
The ones affecting the character directly seem to be fine.  <br>
***
So it seems like migration does not properly check if the item has charges/uses and does not properly choose the right resource to consume.

The amount of uses/charges is properly saved, even if the type is not. Goes for both max/current

It also does not seem to account for adversaries, though I dont know if that is within the scope of this PR.

stanavdb:
@dtfe 

> Debilitate and Work as one both missing resource consumption of focus. Might not be entirely relevant for this PR?

That's not an migration / data issue but ui. See #778 

> All items that consume a resource like uses/charges end up with charges when migration ends.

Fixed.

> Drainer uses charges, not uses. Yet the migration assumes uses for resource consumption, but charges for resources
> Same goes for shardplate

Fixed.

> It also does not seem to account for adversaries, though I dont know if that is within the scope of this PR.

Not an adversary specific issue. Rather, migrations weren't handling migration of actions items directly (only items that would now require embedded actions). Fixed.

dtfe (changes requested):
I'm getting this error after migrating and using any item with a charge or use.
1. Create new character in 2.1
2. Add Drainer
3. Migrate to 3.0
4. Use Drainer in character sheet
5. see error in console:
Image
Note: This does not prevent charges from being consumed, but I'd rather report the error than ignore it.

---

I'm also noting that there is a bit of an issue with how the charges/uses are being migrated on equipment.

It gives each of the items with charges/uses its own action and then assigns those charges/uses to the action, rather than having the charges/uses be on the item itself and then having the action drain it's parent item's charges/uses.

For migrating actions with uses, thats fine. But items should keep the charges/uses on itself.

stanavdb:
> Image 
This error originates from the system embedded collections handlers and is seemingly unrelated.

> I'm also noting that there is a bit of an issue with how the charges/uses are being migrated on equipment.
> 
> It gives each of the items with charges/uses its own action and then assigns those charges/uses to the action, rather than having the charges/uses be on the item itself and then having the action drain it's parent item's charges/uses.
> 
> For migrating actions with uses, thats fine. But items should keep the charges/uses on itself.

Have addressed this. Charges are assigned to the host item now, while uses are still kept on the action. The resource consumption determines where to consume the resource from (`self` or `parent`) based on resource type.

dtfe (changes requested):
So the issue is fixed for charges, not uses.

Example: anesthetic 5 doses item, the uses are on the action still, not the item that holds the doses.

dtfe:
I know you said that you intend for this to be the behavior, but it does not make sense for an item with 5 doses to track the doses on the action of the item, and not the item itself.

dtfe (changes requested):
Almost got it.

Found one issue with this though

I migrated an anesthetic with 3 uses left out of 5
Upon migrating the anesthetic has the required action, and both action and item has 3 uses left out of 5.
If I then go into the anesthetic item, edit the charges up to 5 but keep the action one on 3 and then try to use the action 4 times, it stops after 3 because the action does not have enough uses. Even though the item itself has 2 uses left.

stanavdb:
> Almost got it.
> 
> Found one issue with this though
> 
> I migrated an anesthetic with 3 uses left out of 5 Upon migrating the anesthetic has the required action, and both action and item has 3 uses left out of 5. If I then go into the anesthetic item, edit the charges up to 5 but keep the action one on 3 and then try to use the action 4 times, it stops after 3 because the action does not have enough uses. Even though the item itself has 2 uses left.

Fixed.

dtfe (approved):
Can't find any bugs this time around. Well done

PR #805, opened 2026-07-03
fix(actor): make effects on nested items transfer through to actor

### LINKED ISSUE #800: Effects on embedded actions don't transfer to actor sheets
**Describe the bug**  
Any effects on an embedded action does not transfer to an actor sheet with the embedded action

**To Reproduce**  
1. Create a talent
2. Create an embedded action on that talent
3. Create an effect on that embedded action
4. Make sure that "Apply Effect to Actor" is toggled to true
5. Drag that talent to a new actor
6. Observe that the effect is not propagated to the actor

**Expected behavior**  
Effects should be propagated to the actor

**Foundry VTT Version**  
13.351

|**Name**|**Version**|
|---|---|
|**System**|_3.0,0_|

**Browser**  
VSCode integrated browser

**Additional context**  
_Add any other context about the problem here._

**Checklist:**

stanavdb:
**Type**  
- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
This PR makes transferable effects from items embedded on other items, transfer through to the actor.

**Related Issue**  
Closes #800 

**How Has This Been Tested?**  
1. Create weapon
2. Add effect to weapon
3. Add action to weapon
4. Add effect to action
5. Create actor
6. Add weapon to actor
7. Ensure both effects are present on actor

**Checklist:**  

### DIFF
--- a/src/system/documents/actor.ts
+++ b/src/system/documents/actor.ts
@@ -1346,13 +1346,17 @@ export class CosmereActor<
     }
 
     public *allApplicableEffects() {
-        for (const effect of super.allApplicableEffects()) {
+        for (const effect of this.effects) {
             yield effect;
+        }
+        if (CONFIG.ActiveEffect.legacyTransferral) return;
+        for (const item of this.items) {
+            for (const effect of item.effects) {
+                if (effect.transfer) yield effect;
+            }
 
-            if (effect.parent instanceof CosmereItem) {
-                for (const nestedEffect of effect.parent.nestedEffects) {
-                    if (nestedEffect.transfer) yield nestedEffect;
-                }
+            for (const effect of item.nestedEffects) {
+                if (effect.transfer) yield effect;
             }
         }
     }
--- a/src/system/documents/actor.ts
+++ b/src/system/documents/actor.ts
@@ -1345,17 +1345,17 @@ export class CosmereActor<
         } as const satisfies EnricherData<SubType>;
     }
 
-    // public *allApplicableEffects() {
-    //     for (const effect of super.allApplicableEffects()) {
-    //         if (
-    //             !(effect.parent instanceof CosmereItem) ||
-    //             !effect.parent.isEquippable() ||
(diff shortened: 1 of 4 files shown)


src/system/documents/actor.ts
```diff
@@ -1345,17 +1345,17 @@ export class CosmereActor<
...
+        for (const effect of super.allApplicableEffects()) {
+            yield effect;
+
+            if (effect.parent instanceof CosmereItem) {
+                for (const nestedEffect of effect.parent.nestedEffects) {
+                    if (nestedEffect.transfer) yield nestedEffect;
+                }
+            }
+        }
+    }
```
dtfe, line 1362: This does not work entirely.

super.allApplicableEffects() only detects effects which are 1 embed deep. That means only effects on items directly underneath the actor are recognized. This works with your current test, but does not work if say you have an item with no effect, and an embedded action which does have an effect.

Image

So it will ignore any actions with effects if their parent item does not have an effect. You'll essentially need a for loop that goes through all of the actor's items and checks if they have nested effects.

To reproduce: 
1. Create weapon
2. Add action
3. Give effect to action
4. Give weapon to character
5. See there is no effect in character effects.

dtfe (changes requested):

src/system/documents/actor.ts
dtfe: This should solve the issue
```suggestion
    public *allApplicableEffects() {
        for (const effect of super.allApplicableEffects()){
            yield effect;
        }

        for (const item of this.items){
            for (const nestedEffect of item.nestedEffects){
                if (nestedEffect.transfer) yield nestedEffect;
            }
        }
    }
```

stanavdb: Good spot. I've gone with a slightly different implementation, just to keep the item's effects grouped together. Should be resolved now.

dtfe (approved):
from my testing it seems to work fine now.

PR #814, opened 2026-07-07
Fix #691: avoid recursive mergeObject on live Item instance in rollDa…

### LINKED ISSUE #691: FoundryVTT 14.360 - Attack Rolls don't create a chat message.
**Describe the bug**  
On Foundry VTT 14.360. Whenever performing an attack roll on any actor, no message or roll is created, and `Uncaught (in promise) TypeError: "effects" is read-only` appears in the conole. 

It seems to be called at the `grazeRoll` calculation in `rollDamage`. 

**To Reproduce**  
1. _Go to any actor sheet with an attack action or equipped item_
2. _Click on the item icon to use action_
3. _If using a 'character' and not an adversary, click "Roll"_
4. _See error in dev console_

**Expected behavior**  
An attack roll appears in the chat logs with automated results. 

**Screenshots**  

Image

Image

**Foundry VTT Version**  
14.360

|**Name**|**Version**|
|---|---|
|**System**|2.0.5|

Only modules I have active
Image

**Browser**  
Firefox

**Checklist:**

TheCoderGamer:
**Type**
- [x] Bug fix

**Description**
`rollDamage()` embeds the live `CosmereItem` instance (`this`) inside `rollData.source` via `getDamageRollData()`. The two subsequent `foundry.utils.mergeObject(options, {...})` calls both reuse the same `options` reference, and since `mergeObject` mutates its target and merges nested objects recursively, the second call ends up recursing into the Item instance itself and trying to reassign its properties — including `effects`, which is a read-only getter on `Document` as of Foundry V14. This throws `Cannot assign to read only property 'effects' of object '#<CosmereItem>'` and aborts the roll before any chat message is created

Both calls are replaced with plain object spreads (`{...options, ...}`), which is all that's needed here — a shallow merge of simple keys, never a recursive merge into `source`

**Related Issue**
Closes #691

**How Has This Been Tested?**
- Verified balanced braces/parens in the modified file.
- Reproduce the original error on Foundry V14 with an attack roll, apply the fix, and confirm the attack + damage roll (including graze) now completes and posts to chat normally

**Checklist:**
- [x] I have tested my changes on Foundry VTT version: v14 build 364

### DIFF
--- a/src/system/documents/item.ts
+++ b/src/system/documents/item.ts
@@ -549,18 +549,17 @@ export class CosmereItem<
 
         const formula = options.overrideFormula ?? this.system.damage.formula;
         // Perform the roll
-        const roll = await damageRoll(
-            foundry.utils.mergeObject(options, {
-                formula:
-                    rollData.mod !== undefined
-                        ? `${formula} + ${rollData.mod}`
-                        : formula,
-                damageType: this.system.damage.type,
-                mod: rollData.mod,
-                data: rollData,
-                source: this.name,
-            }) as DamageRollConfiguration,
-        );
+        const roll = await damageRoll({
+            ...options,
+            formula:
+                rollData.mod !== undefined
+                    ? `${formula} + ${rollData.mod}`
+                    : formula,
+            damageType: this.system.damage.type,
+            mod: rollData.mod,
+            data: rollData,
+            source: this.name,
+        } as DamageRollConfiguration);
 
         // Gather the formula options for graze rolls
         const unmoddedRoll = roll.clone();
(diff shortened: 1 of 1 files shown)


Doc-Sun:
@TheCoderGamer While we do appreciate these contributions, we are currently not on, or building for v14. The system is still firmly in v13 and we have a rather large system update coming with release v3.0.0 that will have to be accounted for first. 
Please rest assure that after we have done our 3.0.0 update and finished work on mistborn release we will work on a release for v14 compat and look to getting things like this fixed.

zithith:
In addition to Doc's comments earlier, and to echo my comment on your other v14 PR, once we have released v3.0.0 please merge the new main in and we'll take another look. In this case I suspect there may well be more impact as this upcoming release does involve some significant changes to Item functionality.
We also have a second major update coming to the whole roll engine that's not quite as close to release so it may be worth putting this on hold until after that too.

PR #833, opened 2026-07-13
fix(actor sheet): resolve issue preventing passive effect on compendium actor from being toggled

### LINKED ISSUE #822: Passive Effect on compendium actor unable to be toggled from actor sheet

### Describe the bug

When an effect is embedded on an item that an adversary has, and the adversary is in a compendium (not imported into the world), it is impossible to toggle the effect on the effects tab of the adversary sheet and an error is produced in the console when attempting to toggle the effect.
This is the same fromUuidSync invoked which references an Embedded Document that we've seen in several other places such as #788 

### Expected Behavior

The effect should toggle successfully, or at the very least produce no error in the console

### Steps To Reproduce

1. unlock starter rules Companions and Adversaries compendium for editing
2. open up any adversary, I tested with the Archer
3. edit one of the adversary's features
4. add a toggleable effect to the feature in the Effects tab
5. attempt to toggle the effect on the actor sheet
6. view error in console

### Operating System

Windows 11

### Web Browser(s)

Foundry Executable

### Hosting Method

Localhost/same PC

13.351

3.0.0

Image

dtfe:
**Type**  

- [X] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
Fixes issue where FromUuidSync is used to get effects on compendium items which does not work as intended. Replaced code with a new method to get an effect from a UUID directly from the actor itself which works for compendium actors.

**Related Issue**  
Closes #822 

**How Has This Been Tested?**  
1. Unlock companions and adversaries in starter rules
2. Open it and go to companion, special, greater larkin
3. go to effects and toggle effect and see it gets toggled.

**Checklist:**  

### DIFF
--- a/src/system/documents/actor.ts
+++ b/src/system/documents/actor.ts
@@ -624,12 +624,18 @@ export class CosmereActor<
 
     /** Returns an embedded item in an actor regardless of how deeply nested it is, if it exists.
      *
-     * @param itemUuid The UUID of the item you want to get
+     * @param itemUuid The UUID of the item you want to get, either the whole UUID or partial. Also works with just the item ID.
      * @returns An Item or null if no item could be found.
      */
     public getNestedEmbeddedItemFromUuid(itemUuid: string): Item | null {
+        const parsedUuid = foundry.utils.parseUuid(itemUuid);
+        let itemId = parsedUuid?.id;
+        if (!itemId) itemId = itemUuid;
         for (const [, document] of this.traverseEmbeddedDocuments()) {
-            if (document instanceof Item && document.uuid === itemUuid) {
+            if (
+                document instanceof Item &&
+                (document.uuid === itemUuid || document.id == itemId)
+            ) {
                 return document;
             }
         }
--- a/src/system/documents/actor.ts
+++ b/src/system/documents/actor.ts
@@ -622,41 +622,22 @@ export class CosmereActor<
 
     /* --- Functions --- */
 
-    /** Returns an embedded item in an actor regardless of how deeply nested it is, if it exists.
+    /** Returns an embedded document in an actor regardless of how deeply nested it is, if it exists.
      *
-     * @param itemUuid The UUID of the item you want to get, either the whole UUID or partial. Also works with just the item ID.
-     * @returns An Item or null if no item could be found.
+     * @param documentType the type of document you are getting. In this instance thats either {@link CosmereActiveEffect} or {@link CosmereItem}
+     * @param uuid The UUID of the document you want to get, either the whole UUID or partial. Also works with just the document ID.
+     * @returns A {@link CosmereActiveEffect}, {@link CosmereItem} or null if no effect could be found.
      */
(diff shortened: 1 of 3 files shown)


stanavdb (commented):
Is there a reason we couldn't just use `fromUuid` outright here, instead of building custom search? Following the path up from `getEffectFromElement` leads to handler functions which are all async.

dtfe:
> Is there a reason we couldn't just use `fromUuid` outright here, instead of building custom search? Following the path up from `getEffectFromElement` leads to handler functions which are all async.

The main reason is that I did not want to potentially introduce bugs by modifying multiple functions to be asynchronous. Especially this close to the deadline.

It is also used by the context menu which if that isnt asynchronous and expects synchronous behavior then that would potentially cause issues. Seems safest to use this implementation.

src/system/documents/actor.ts
```diff
@@ -624,19 +624,46 @@ export class CosmereActor<
...
+        for (const [, document] of this.traverseEmbeddedDocuments()) {
+            if (
+                document instanceof CosmereActiveEffect &&
+                (document.uuid == effectUuid || document.id == effectId)
+            ) {
+                return document;
+            }
+        }
+        return null;
+    }
```
stanavdb: We now have code duplication as we have two fairly similar functions that do just slightly different things. These should be combined into a single function that can retrieve any embedded document.

stanavdb (changes requested):
We can stick with custom search function for now, given the above reasoning. Can approve once code duplication has been addressed.

src/system/documents/actor.ts
dtfe: Good call. For some reason I thought it would be more efficient to separate them, as if traverseEmbeddedDocuments knew to only check effects or only items, but that's not how it works.

src/system/documents/actor.ts
```diff
@@ -622,41 +622,22 @@ export class CosmereActor<
...
+     * @param documentType The type of document you are getting. In this instance thats either {@link CosmereActiveEffect} or {@link CosmereItem}
+     * @param uuid The UUID of the document you want to get, either the whole UUID or partial. Also works with just the document ID.
+     * @returns A {@link CosmereActiveEffect}, {@link CosmereItem} or null if no document could be found.
      */
-    public getNestedEmbeddedItemFromUuid(itemUuid: string): Item | null {
-        const parsedUuid = foundry.utils.parseUuid(itemUuid);
-        let itemId = parsedUuid?.id;
-        if (!itemId) itemId = itemUuid;
+    public getEmbeddedDocumentFromUuid(
+        documentType: typeof CosmereActiveEffect | typeof CosmereItem,
```
stanavdb: You can get the document type from `foundry.utils.parseUuid` (`documentType` or `type` fields) and compare to the `.documentName` field on any document

stanavdb (changes requested):

stanavdb (approved):

PR #845, opened 2026-07-14
fix(actor sheet): Fixes issue where certain events are fired twice

### LINKED ISSUE #844: Actions duplicate when used

### Describe the bug

Title

### Expected Behavior

Fires once, not twice.

### Steps To Reproduce

1. create character
2. create action
3. use action
4. See it is fired twice when used

### Operating System

Windows 11

### Web Browser(s)

Foundry Executable

### Hosting Method

Localhost/same PC

13.351

3.0.0

_No response_

dtfe:
**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
Fixes issue where the action list events are fired twice.

**Related Issue**  
Closes #844 

**How Has This Been Tested?**  
1. Open new character sheet
2. Go to actions
3. Add new action and see that it adds a new action
4. Try to use new action and it should only be used once
5. Try to use the edit button and see it works
6. Try to expand and contract and see it works with no errors
7. try to contract and expand the action section itself
8. Add a radiant path
9. See that you can do the same for talents as steps 5-7

**Checklist:**  

### DIFF
--- a/src/system/applications/actor/components/actions-list-entry.ts
+++ b/src/system/applications/actor/components/actions-list-entry.ts
@@ -25,7 +25,6 @@ export class ActionsListEntryComponent extends ActorItemListComponent {
      */
     /* eslint-disable @typescript-eslint/unbound-method */
     static readonly ACTIONS = {
-        ...super.ACTIONS,
         'toggle-action-details': this.onToggleActionDetails,
         'use-item': this.onUseItem,
     };
--- a/src/system/applications/actor/components/actions-list-entry.ts
+++ b/src/system/applications/actor/components/actions-list-entry.ts
@@ -25,7 +25,7 @@ export class ActionsListEntryComponent extends ActorItemListComponent {
      */
     /* eslint-disable @typescript-eslint/unbound-method */
     static readonly ACTIONS = {
-        'toggle-action-details': this.onToggleActionDetails,
+        ...super.ACTIONS,
         'use-item': this.onUseItem,
     };
     /* eslint-enable @typescript-eslint/unbound-method */
--- a/src/system/applications/actor/components/actions-list.ts
+++ b/src/system/applications/actor/components/actions-list.ts
@@ -257,9 +257,11 @@ export class ActorActionsListComponent extends ActorItemListComponent {
      * NOTE: Unbound methods is the standard for defining actions
      * within ApplicationV2
      */
-
+    /* eslint-disable @typescript-eslint/unbound-method */
     static readonly ACTIONS = {
         ...super.ACTIONS,
+        'toggle-section-collapsed': this.onToggleSectionCollapsed,
+        'new-item': this.onNewItem,
     };
 
     /* --- Context --- */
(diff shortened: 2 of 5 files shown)


src/system/applications/actor/components/item-list.ts
```diff
@@ -34,19 +34,6 @@ export class ActorItemListComponent extends HandlebarsApplicationComponent<// ty
...
-    /**
-     * NOTE: Unbound methods is the standard for defining actions
-     * within ApplicationV2
-     */
-    /* eslint-disable @typescript-eslint/unbound-method */
-    static readonly ACTIONS = {
-        'toggle-section-collapsed': this.onToggleSectionCollapsed,
-        'toggle-action-details': this.onToggleActionDetails,
-        'use-item': this.onUseItem,
-        'new-item': this.onNewItem,
```
posadist-revolution, line 46: We should remove the "use-item" and "new-item", but toggle-section-collapsed and toggle-action-details should both live here.

posadist-revolution: Actually, on second thought, new-item should live here as well, as that's common behavior for essentially all collapsible lists on the character sheet. Only use-item should be removed.

src/system/applications/actor/components/actions-list-entry.ts
```diff
@@ -25,7 +25,6 @@ export class ActionsListEntryComponent extends ActorItemListComponent {
      */
     /* eslint-disable @typescript-eslint/unbound-method */
     static readonly ACTIONS = {
-        ...super.ACTIONS,
         'toggle-action-details': this.onToggleActionDetails,
```
posadist-revolution: With my comment on item-list.ts, we should keep ...super.ACTIONS but remove toggle-action-details here.

src/system/applications/actor/components/actions-list.ts
```diff
@@ -257,9 +257,11 @@ export class ActorActionsListComponent extends ActorItemListComponent {
      * NOTE: Unbound methods is the standard for defining actions
      * within ApplicationV2
      */
-
+    /* eslint-disable @typescript-eslint/unbound-method */
     static readonly ACTIONS = {
         ...super.ACTIONS,
+        'toggle-section-collapsed': this.onToggleSectionCollapsed,
+        'new-item': this.onNewItem,
```
posadist-revolution: Both of these should not be necessary with my suggestions for item-list.ts

posadist-revolution (changes requested):

src/system/applications/actor/components/effects-list.ts
```diff
@@ -45,7 +45,6 @@ export class ActorEffectsListComponent extends ActorItemListComponent {
     /* eslint-disable @typescript-eslint/unbound-method */
     static readonly ACTIONS = {
         ...super.ACTIONS,
-        'toggle-action-details': this.onToggleActionDetails,
         'toggle-section-collapsed': this.onToggleSectionCollapsed,
```
posadist-revolution, line 48: We probably don't need this one either, since it already comes from `...super.ACTIONS`

posadist-revolution (changes requested):

src/system/applications/actor/components/effects-list.ts
dtfe: No, that one is necessary as effects-list has its own version of that method so it has to override super.ACTIONS.

posadist-revolution: Just tested and confirmed you're correct- I was misunderstanding how object-oriented stuff works with static because I've spent too much time in an embedded space 😆

posadist-revolution (approved):

Issue #850, opened 2026-07-15
Embedded actions sheets don't re-render after certain updates on 3.0.0


stanavdb:
### Is there an existing issue for this?

- [x] I have searched the existing issues to ensure this bug has not already been reported

### Describe the bug

All document updates to embedded actions should cause their sheet to get re-rendered. However after performing two updates in a row, this doesn't happen, leaving the sheet stale. Closing and re-opening the sheet shows the correct state.

### Expected Behavior

Sheet should always match the state of the data.

### Steps To Reproduce

1. Create a weapon
2. Add an action to the weapon
3. Add an event to the action (first update)
4. Delete the event (second update)

### Operating System

Windows 11

### Web Browser(s)

Foundry Executable

### Hosting Method

Localhost/same PC

### Version Info - Foundry VTT

13.351

### Version Info - Cosmere RPG

3.0.0

Issue #841, opened 2026-07-14
Embedded actions do not save when creating a new adventure


dtfe:
### Is there an existing issue for this?

- [x] I have searched the existing issues to ensure this bug has not already been reported

### Describe the bug

Rebuilding an adventure and then importing it keeps the embedded action, but when creating a new adventure (Not rebuild) the embedded actions are not included.

### Expected Behavior

That it saves all the data needed, including embedded actions

### Steps To Reproduce

1. Create a new world
2. Create new item
3. add action to item
5. Create new compendium for adventures
6. add new adventure to compendium
7. add action to adventure contents
8. build
9. delete item from world
10. import adventure
11. open item and see it lacks the action

### Operating System

Windows 11

### Web Browser(s)

Foundry Executable

### Hosting Method

Localhost/same PC

### Version Info - Foundry VTT

13.351

### Version Info - Cosmere RPG

3.0.0

### Screenshots and Additional Context

Related to #840 and #839

PR #854, opened 2026-07-16
fix(system embedded documents): add special handling for adventures

### LINKED ISSUE #839: Adventures do not correctly handle system embedded documents on 3.0

### Describe the bug

Building an adventure with items that contain embedded items (system embedded documents) does not save those embedded items. Importing an adventure that was manually updated to contain items with embedded items doesn't bring those deeply embedded items in.

### Expected Behavior

Adventures should handle system embedded documents.

### Steps To Reproduce

1. Prepare compendium with an adventure
2. Create an item
3. Add an action to the item
4. Go to adventure, right click and select "rebuild adventure"
5. Go to contents and add the item
6. Click build adventure
7. Delete the item from the world
8. Import the adventure
9. Verify that the action is now missing

### Operating System

Windows 11

### Web Browser(s)

Foundry Executable

### Hosting Method

Localhost/same PC

13.351

3.0.0

_No response_

stanavdb:
**Type**  
- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
Fixes issues with the system embedded documents handling of adventures. This time without breaking actor sheets / events.

**Related Issue**  
Closes #839 

**How Has This Been Tested?**  
1. Prepare compendium with an adventure
2. Create an item
3. Add an action to the item
4. Go to adventure, right click and select "rebuild adventure"
5. Go to contents and add the item
6. Click build adventure
7. Delete the item from the world
8. Import the adventure
9. Verify that the action is present

**Checklist:**  

### DIFF
--- a/src/system/documents/system-embedded-collections/utils/socket.ts
+++ b/src/system/documents/system-embedded-collections/utils/socket.ts
@@ -607,22 +607,16 @@ function transformCreateUpdateResponse(
                         ),
                 );
         } else {
-            const documentCls = CONFIG[inRequest.type]?.documentClass as
-                | Document.Constructable.AnyConstructor
-                | undefined;
-
-            if (documentCls && hasSystemEmbeddedCollections(documentCls)) {
-                result = result.map((doc) =>
-                    toClientViewObject(doc as AnyObject, inRequest.type),
-                );
-            }
+            result = result.map((doc) =>
+                toClientViewObject(doc as AnyObject, inRequest.type),
+            );
         }
 
         const targetIds = inResponse.operation.targets?.map(
             (target) => target.id,
         );
         if (targetIds)
-            result = (result as AnyObject[]).filter((doc) =>
+            result = result.filter((doc) =>
                 targetIds.includes(
                     foundry.utils.getProperty(doc, '_id') as string,
                 ),
--- a/src/system/documents/system-embedded-collections/utils/socket.ts
+++ b/src/system/documents/system-embedded-collections/utils/socket.ts
@@ -144,37 +144,30 @@ async function transformCRUDRequest(
 
     if (!hierarchy.includesSystemEmbedding || !hierarchy.host) {
         if (isCreateRequest(inRequest)) {
-            return foundry.utils.mergeObject(
-                transformRequestCommon(inRequest),
-                {
-                    operation: {
-                        data: inRequest.operation.data.map((data) =>
-                            data
-                                ? toServerViewObject(
-                                      data as AnyDocumentData,
-                                      inRequest.type,
-                                  )
-                                : data,
-                        ),
-                    },
-                },
+            const outRequest = transformRequestCommon(inRequest);
+            outRequest.operation.data = inRequest.operation.data.map((data) =>
+                data
+                    ? toServerViewObject(
+                          data as AnyDocumentData,
+                          inRequest.type,
+                      )
+                    : data,
             );
+
+            return outRequest;
         } else if (isUpdateRequest(inRequest)) {
-            return foundry.utils.mergeObject(
-                transformRequestCommon(inRequest),
-                {
-                    operation: {
-                        updates: inRequest.operation.updates.map((data) =>
-                            data
-                                ? toServerViewObject(
-                                      data as AnyDocumentData,
-                                      inRequest.type,
-                                  )
-                                : data,
-                        ),
-                    },
-                },
+            const outRequest = transformRequestCommon(inRequest);
+            outRequest.operation.updates = inRequest.operation.updates.map(
+                (data) =>
+                    data
+                        ? toServerViewObject(
+                              data as AnyDocumentData,
+                              inRequest.type,
+                          )
+                        : data,
             );
+
+            return outRequest;
         } else {
             return inRequest;
         }
@@ -635,9 +628,10 @@ function transformCreateUpdateResponse(
             );
     }
 
-    return foundry.utils.mergeObject(transformCRUDResponseCommon(inResponse), {
-        result,
-    });
+    const outResponse = transformCRUDResponseCommon(inResponse);
+    outResponse.result = result;
+
+    return outResponse;
 }
 
 function transformDeleteResponse(inResponse: SocketResponse): SocketResponse {
--- a/src/system/documents/system-embedded-collections/utils/socket.ts
+++ b/src/system/documents/system-embedded-collections/utils/socket.ts
@@ -211,6 +211,7 @@ async function transformCRUDRequest(
                         id: doc.id,
                     })),
                 queue: true,
+                isHierarchical: true,
             },
         };
 
@@ -667,7 +668,9 @@ function transformCRUDResponseCommon(
             parentUuid: inRequest.operation.parentUuid,
             render: inRequest.operation.render,
             diff: false,
-            recursive: false,
+            recursive: inRequest.operation.isHierarchical
+                ? true
+                : inRequest.operation.recursive,
             renderSheet: foundry.utils.getProperty(
                 inRequest.operation,
                 'renderSheet',
--- a/src/system/documents/system-embedded-collections/utils/socket.ts
+++ b/src/system/documents/system-embedded-collections/utils/socket.ts
@@ -700,7 +700,7 @@ function transformAdventureResponse(
             const data = foundry.utils.getProperty(r, key) as AnyObject[];
 
             foundry.utils.setProperty(
-                resolveUpdate,
+                r,
                 key,
                 data.map((data) => toClientViewObject(data, documentName)),
             );
--- a/src/system/documents/system-embedded-collections/utils/socket.ts
+++ b/src/system/documents/system-embedded-collections/utils/socket.ts
@@ -502,10 +502,15 @@ export function transformResponse(inResponse: SocketResponse): SocketResponse {
 
     if (isGetRequest(inRequest)) {
         if (inRequest.type === 'Adventure')
-            return transformAdventureGetResponse(inResponse);
+            return transformAdventureResponse(inResponse);
 
         return transformGetResponse(inResponse);
-    } else if (isCreateRequest(inRequest) || isUpdateRequest(inRequest)) {
+    } else if (isCreateRequest(inRequest)) {
+        return transformCreateUpdateResponse(inResponse);
+    } else if (isUpdateRequest(inRequest)) {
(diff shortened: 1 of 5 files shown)


dtfe (changes requested):
Embedded action's events are not live updated anymore, so when pressing + it does not seem to add a new event. Same goes for deleting. Can update them by editing them though.
Seems like this lack of live updates also applies to setting a resource consumption and resource on the action.

stanavdb:
> Embedded action's events are not live updated anymore, so when pressing + it does not seem to add a new event. Same goes for deleting. Can update them by editing them though. Seems like this lack of live updates also applies to setting a resource consumption and resource on the action.

This issue isn't caused by this PR and is present on release 3.0.0, see #850

dtfe:
Im going to hold off on reviewing this until #850 is merged and this has been updated. It'll be very hard to test relevant things in this PR without the live updates on embedded actions.

src/system/documents/system-embedded-collections/utils/socket.ts
```diff
@@ -578,25 +655,65 @@ function transformCRUDResponseCommon(
...
+            const documentName = (
+                (field as foundry.data.fields.SetField.Any)
+                    .element as foundry.data.fields.EmbeddedDataField<
+                    typeof foundry.abstract.Document
+                >
+            ).model.documentName;
+            const data = foundry.utils.getProperty(r, key) as AnyObject[];
+
+            foundry.utils.setProperty(
+                resolveUpdate,
```
dtfe: This should probably be "r" not "resolveUpdate"

Don't think r is changed at all during this map otherwise.

src/system/documents/system-embedded-collections/utils/socket.ts
dtfe, line 101: Are these supposed to be different? Could this cause issues? Cause it seems like it takes a CRUDAction but only returns updates, which suggests it lacks create and delete?

dtfe, line 859: Same as above, should there be an error thrown if it can't find a documentClass? The rest of the function relies on cls being valid, I think.

dtfe, line 743: Should there be an error thrown if it can't find a documentClass?

dtfe, line 285: Like above, I recommend a switch here for clarity, unless theres a reason you aren't.

src/system/documents/system-embedded-collections/utils/socket.ts
```diff
@@ -105,6 +105,14 @@ export function transformRequest(
...
     inRequest: DocumentSocketRequest,
 ): Promise<DocumentSocketRequest> {
+    if (inRequest.type === 'Adventure') {
+        if (isGetRequest(inRequest)) {
+            return transformAdventureGetRequest(inRequest);
+        } else if (isUpdateRequest(inRequest)) {
+            return transformAdventureUpdateRequest(inRequest);
+        }
+    }
+
```
dtfe, line 115: This only handles get and update requests. Is this the reason why the bug this aims to solve remains when you create a new adventure instead of rebuilding?

Also could we use a switch here?
From what I can tell "isBlankRequest()" essentially just returns the inRequest.action variable which means it could probably just be put into a switch and be a lot clearer rather than chaining if/else

src/system/documents/system-embedded-collections/utils/socket.ts
```diff
@@ -135,20 +143,34 @@ async function transformCRUDRequest(
...
                         : data,
-                ),
-            },
-        });
+            );
+
+            return outRequest;
+        } else {
+            return inRequest;
+        }
```
dtfe, line 173: Could this also use a switch?

src/system/documents/system-embedded-collections/utils/socket.ts
```diff
@@ -395,6 +419,50 @@ function resolveUpdate(
...
+        embeddedDataFields.forEach(([key, field]) => {
+            if (!foundry.utils.hasProperty(update, key)) return;
+
+            const documentName = (
+                (field as foundry.data.fields.SetField.Any)
+                    .element as foundry.data.fields.EmbeddedDataField<
+                    typeof foundry.abstract.Document
+                >
+            ).model.documentName;
+            const data = foundry.utils.getProperty(update, key) as AnyObject[];
```
dtfe, line 451: In other areas of the code you use the method "ensureArray" but here you assume there is an array. Is there a reason for this?

src/system/documents/system-embedded-collections/utils/socket.ts
```diff
@@ -395,6 +419,50 @@ function resolveUpdate(
...
 
+function transformAdventureGetRequest(
+    inRequest: DocumentSocketRequest<'get'>,
+): DocumentSocketRequest<'get'> {
+    return transformRequestCommon(inRequest);
+}
+
+function transformAdventureUpdateRequest(
+    inRequest: DocumentSocketRequest<'update'>,
+): DocumentSocketRequest {
```
dtfe, line 430: Is there a reason why this returns a normal DocumentSocketRequest and now 'update'?

src/system/documents/system-embedded-collections/utils/socket.ts
dtfe, line 515: I'd make this a switch unless theres a reason not to.

dtfe (changes requested):
So if you have an character or adversary with an item that has an embedded action in the compendium, that item loses its embedded action and thus the actor also loses it. Aside from that I couldnt find any other issues. 

I've also left a good few comments on the socket.ts script. Decided to sit down and try to read and understand the code so I left a few questions on the lines I found odd.

src/system/documents/system-embedded-collections/utils/socket.ts
stanavdb: In practice there is no scenario where `CONFIG[documentType]` isn't defined. This is a case of foundry-vtt-types being aggressive in its correctness.

src/system/documents/system-embedded-collections/utils/socket.ts
stanavdb: Good catch, that's a bug yeah

src/system/documents/system-embedded-collections/utils/socket.ts
stanavdb: Can't use a switch case here as these `is...Request` functions serve as typeguards. If we just check the action, we'll lack the request type definition.

src/system/documents/system-embedded-collections/utils/socket.ts
stanavdb: `ensureArray` is only used where we're accessing document embedded collection fields. Here we are accessing the update value of a set field that carries embedded data, which has gone through Foundry's data validation pipeline so we know it's an array.

src/system/documents/system-embedded-collections/utils/socket.ts
stanavdb: A "normal" `DocumentSocketRequest` is just any of `create`, `update`, `delete`, `get`. This can be typed with `update` but it's not necessary since we're not accessing the type later anymore.  
No real reason it was done this way over the other.

src/system/documents/system-embedded-collections/utils/socket.ts
stanavdb: See my other reply regarding switch

src/system/documents/system-embedded-collections/utils/socket.ts
stanavdb: See my other reply regarding switch

src/system/documents/system-embedded-collections/utils/socket.ts
stanavdb: > This only handles get and update requests. Is this the reason why the bug this aims to solve remains when you create a new adventure instead of rebuilding?

Likely, though haven't confirmed.

src/system/documents/system-embedded-collections/utils/socket.ts
stanavdb: Yes, on the client side we have actual embedded documents for system defined embeds (i.e. items on items). However that doesn't exist on the backend (and can't, since system code ONLY exists client side). On the backend we store those embedded documents under the `system` field (specifically `system.__embedded`). So any create and delete operations are really update operations on the parent.

dtfe (changes requested):
I forgot to leave the steps for the bug I reported.
1. Create new compendium - adventure
2. Create new adventure
3. build
4. create new talent
5. add embedded action to talent
6. create new character
7. add talent to character
8. create new adversary feature
9. Add embedded action to feature
10. create new adversary
11. add feature to adversary
12. rebuild adventure
13. add character and adversary to adventure contents
14. build
15. delete character/adversary
16. import adventure
17. see actors are missing embedded actions from their respective talent/feature.

stanavdb:
> I forgot to leave the steps for the bug I reported.
> 
>     1. Create new compendium - adventure
> 
>     2. Create new adventure
> 
>     3. build
> 
>     4. create new talent
> 
>     5. add embedded action to talent
> 
>     6. create new character
> 
>     7. add talent to character
> 
>     8. create new adversary feature
> 
>     9. Add embedded action to feature
> 
>     10. create new adversary
> 
>     11. add feature to adversary
> 
>     12. rebuild adventure
> 
>     13. add character and adversary to adventure contents
> 
>     14. build
> 
>     15. delete character/adversary
> 
>     16. import adventure
> 
>     17. see actors are missing embedded actions from their respective talent/feature.

@dtfe This one is now fixed and was actually a bigger underlying issue that would've likely caused other problems with/around actors (or any other document that doesn't itself contain system embedded collections but does embed documents that do).

dtfe (approved):
I can't find any mechanical issues doing my tests this time around so hopefully we've squashed most of the bugs.

Here's the things I tested:

**Weapon**
- Weapons with embedded strikes have the right values and import fine.

**Character**
- Talent trees import fine
- Talents with embedded actions import fine
- Talents with embedded actions within talent trees import fine.
- Actors with a talent with embedded actions import fine.
- Actors with a power that contains a talent tree that contains a talent that contains an embedded action import fine.

**Adversary**
- Adversary features with embedded actions import fine. 
- Adversaries with adversary features with features with embedded actions import fine.

**Token**
- Actor tokens move as they should
- Targeting works and displays the correct defense values

The only remaining issue isn't super critical and has an issue for it already #841 
So this seems good to go.

PR #858, opened 2026-07-18
feat(character sheet): add item resource consumption 

### LINKED ISSUE #778: Integrate item resources into "Cost" section of character sheet Actions tab

posadist-revolution:
**Type**  

- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
Adds the ability to display item resource costs on the character sheet "Actions" tab, with a shortened version of the target link. The link also is able to display an error icon when the item with resources can't be found on the sheet.

**Related Issue**  
Closes #778 

**How Has This Been Tested?**  
1. Created a new action on an actor sheet. 
2. Set up a new equippable item named "Test" in a new compendium, with a "charges" resource
3. Configured the action on the actor sheet to consume 1 charge from "test".
4. Verified that the actions display now includes "1 charge from [ERROR SYMBOL] [test icon]", with a link directing to the "Test" item reference in the compendium
5. Added "Test" to the character
6. Verified that the actions display did not change
7. Equipped "Test" on the actor
8. Verified that the actions display error symbol disappeared, and the link directed to the copy of "test" on the actor sheet.

**Checklist:**  

### DIFF
--- a/src/system/utils/handlebars/index.ts
+++ b/src/system/utils/handlebars/index.ts
@@ -545,63 +545,14 @@ Handlebars.registerHelper(
     'resourceCostLabel',
     (consume: ActionItemDataModel.ConsumeData) => {
         const { value } = consume;
+        let resource = '';
         if (consume.type === ItemConsumeType.Resource) {
-            const resource = game.i18n.localize(
+            resource = game.i18n.localize(
                 CONFIG.COSMERE.resources[consume.resource].label ??
                     'GENERIC.Unknown',
             );
-
-            let label = '';
-
-            // Get adjusted minimum value, to account for optional formatting
-            const adjustedMin = Math.max(value.min, 1);
-
-            // Static range
-            if (adjustedMin === value.max) {
-                label = game.i18n.format(
-                    'COSMERE.Actor.Sheet.Actions.Consume.Static',
-                    {
-                        amount: adjustedMin.toFixed(),
-                        resource,
-                    },
-                );
-            }
-            // Uncapped range
-            else if (value.max === -1) {
-                label = game.i18n.format(
-                    'COSMERE.Actor.Sheet.Actions.Consume.RangeUncapped',
-                    {
-                        amount: adjustedMin.toFixed(),
-                        resource,
-                    },
-                );
-            }
-            // Capped range
-            else {
-                label = game.i18n.format(
-                    'COSMERE.Actor.Sheet.Actions.Consume.RangeCapped',
-                    {
-                        min: adjustedMin.toFixed(),
-                        max: value.max.toFixed(),
-                        resource,
-                    },
-                );
-            }
-
-            // Treat actual minimum value of 0 as an "optional" cost
-            if (value.min === 0) {
-                label = game.i18n.format(
-                    'COSMERE.Actor.Sheet.Actions.Consume.Optional',
-                    {
-                        label,
-                    },
-                );
-            }
-
-            return label;
         } else if (consume.type === ItemConsumeType.ItemResource) {
             const singular = value.min == 1 && value.max == 1;
-            let resource = '';
             if (singular) {
                 resource = game.i18n.localize(
                     CONFIG.COSMERE.item.resource.types[consume.resource]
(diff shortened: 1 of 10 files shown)


Doc-Sun (commented):
Usage Tested, needs full code review

src/templates/general/components/match-document-target.hbs
```diff
@@ -46,10 +86,11 @@
...
 {{#if editable}}
     <div class="controls">
-        <a data-action="configure" 
+        <a data-action="configure"
             data-tooltip="COSMERE.Utils.MatchDocument.Target.Configure"
         >
             <i class="fa-solid fa-gear"></i>
         </a>
     </div>
+{{/if}}
```
dtfe, line 96: Don't think this /if is needed

dtfe (changes requested):
Aside from this one extra /if, I don't think there's any issues with the code.

But you may want to get a third opinion from someone more used with handlebars

src/templates/general/components/match-document-target.hbs
```diff
@@ -46,10 +86,11 @@
...
 {{/if}}
```
dtfe, line 97: Sorry, didn't select both.

src/templates/general/components/match-document-target.hbs
posadist-revolution: It actually is- this is the ending of the {{else}} block for the long-form version of the target reference.

src/system/utils/handlebars/index.ts
```diff
@@ -530,62 +545,127 @@ Handlebars.registerHelper(
...
+
+            // Treat actual minimum value of 0 as an "optional" cost
+            if (value.min === 0) {
+                label = game.i18n.format(
+                    'COSMERE.Actor.Sheet.Actions.Consume.Optional',
+                    {
+                        label,
+                    },
+                );
+            }
```
stanavdb: You've got a lot of code duplication here while the if is essentially just determining the resource label

src/templates/actors/components/actions-list-entry.hbs
```diff
@@ -48,7 +48,19 @@
                     {{#if consume.consumesResource}}
                     <span>{{resourceCostLabel consume}}</span>
                     {{/if}}
+                    {{#if consume.consumesItemResource}}
 
+                    <span>{{resourceCostLabel consume}}</span>
+
+                    <span>from</span>
```
stanavdb: Needs localization

src/templates/general/components/match-document-target.hbs
```diff
@@ -1,9 +1,49 @@
...
+    >
+        {{#if resolvedDocument.img}}
+            <img src="{{resolvedDocument.img}}" alt="{{resolvedDocument.name}}">
+        {{/if}}
+    </a>
+
+{{else if (gt value.steps.length 0)}}
+    {{#with (last value.steps) as |step|}}
+        {{#if (not (or (eq step.target "self") (eq step.target "parent")))}}
+            {{#if (and (eq step.matchBy "document-type") step.documentType)}}
```
stanavdb, line 19: Was the `<span class="type">{{localize (concat "DOCUMENT." step.documentType)}}</span>` line intentionally left out here?

src/templates/general/components/match-document-target.hbs
stanavdb, line 1: By splitting short / long at the root of the file you're introducing a lot of code duplication.

stanavdb (changes requested):

src/templates/general/components/match-document-target.hbs
posadist-revolution: Yes- I'm not in love with that, but to avoid duplication we'd need to put if statements scattered throughout the file in a variety of places to make the changes we'd like to make. With the tradeoff being either "Duplicate code and have a much simpler and more understandable file" vs "Not duplicate code but substantially increase the logic complexity of the file" I decided to go with the code duplication, especially since this isn't a particularly long file in any case.

src/templates/general/components/match-document-target.hbs
posadist-revolution: That was a mistake, thanks for catching it! Fixed now.

src/templates/actors/components/actions-list-entry.hbs
posadist-revolution: Fixed

src/system/utils/handlebars/index.ts
posadist-revolution: Fixed

src/templates/general/components/match-document-target.hbs
stanavdb: Don't fully agree with the sentiment that code duplication makes this easier to read, or that instead of addressing the readability properly we it introduce code duplication. However, since this will all get refactored out anyway, I'm going to approve.

stanavdb (approved):

dtfe (approved):

PR #867, opened 2026-07-19
feat(actor sheets): make consume section on actions show current charges

### LINKED ISSUE #868: Easily display resources available/max on resource consumption on action lists

- [x] I have searched the existing issues to ensure this feature has not already been requested

### Is your feature request related to a problem? Please describe.

Action list resource consumption doesn't always easily show remaining resources

### Describe the solution you'd like.

Show remaining resources

### Describe alternatives you've considered

Not showing them

3.0.0

_No response_

posadist-revolution:
**Type**  

- [ ] Bug fix
- [x] Feature
- [ ] Refactor

**Description**  
Added a display for "current/max" resource on a "cost" section of an actor sheet

**Related Issue**  
Closes #868 

**How Has This Been Tested?**  
Created an actor with an equippable item with charges (1/3), added an action which consumes 1 charge from that item, verified that the resulting display shows (1/3) before the icon
Added Uses (2/4) to that same item, created another action which consumes 1 use from that item, verified that the resulting display shows (2/4) before the icon

**Checklist:**  

### DIFF
--- a/src/system/applications/components/match-document-target.ts
+++ b/src/system/applications/components/match-document-target.ts
@@ -17,8 +17,6 @@ import DataModelUtils from '@system/utils/data-model';
 // Constants
 import { SYSTEM_ID } from '@src/system/constants';
 import { TEMPLATES } from '@src/system/utils/templates';
-import { ItemResource } from '@src/system/types/cosmere';
-import { CosmereItem } from '@src/system/documents';
 
 export class MatchDocumentTargetComponent extends HandlebarsApplicationComponent<
     foundry.applications.api.ApplicationV2.AnyConstructor,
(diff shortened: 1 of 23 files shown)


src/system/applications/components/match-document-target.ts
```diff
@@ -138,10 +140,25 @@ export class MatchDocumentTargetComponent extends HandlebarsApplicationComponent
             typeof lastStep.reference === 'string'
                 ? await fromUuid(lastStep.reference)
                 : lastStep.reference;
+        let resolvedDocumentResourceValue;
+        let resolvedDocumentResourceMax;
+        if (
+            resolvedDocument &&
+            params.resourceId &&
+            (resolvedDocument as CosmereItem).hasResources()
```
stanavdb: This is not a safe type assertion. The component is built generically to be able to target any document. Use a type guard instead.

src/system/applications/components/match-document-target.ts
stanavdb, line 1: I don't agree with the approach of adding resource info to the `match-document-target` component; The component has nothing to do with resources otherwise, nor does the match document logic inherently.

stanavdb (changes requested):

src/system/applications/components/match-document-target.ts
posadist-revolution: I think that's fair. This was imagined mostly as a stopgap change, with the longer-term change being a way to show this data in the "Resources" section of the relevant action. 
The reason I put this up instead of a change to the "resources" section is that before we do that change, we need to create a way to indicate to users that "This resource is relevant for this action, but does not actually live on this particular action" distinctly from "This is a resource that lives on this individual action".
Are you okay with this as a stopgap change that we revert once we have a better way to tie into the "resources" section?

stanavdb: Every change we put in, we should assume is permanent, since there's always going to be a higher prio issue to take care of due to our limited time. That's how we ended up with the mess that is the current roll workflow.
  
Aside from that, there's no reason to modify the `match-document-target` component for this; You could use a wrapper component that uses the `match-document-target` and just adds the resources overlay.

posadist-revolution: Makes sense and I think a wrapper around it is a reasonable solution. I'll change this PR to target that instead, hopefully tonight!

stanavdb (approved):

PR #874, opened 2026-07-24
feat(actor sheets): add ability for users to post actor talents and features to chat

### LINKED ISSUE #769: Talents should be postable to chat
After the action refactor Talent items on the character sheet are no longer clickable. This used to post them to chat so they could be read by anyone, even if they had no action associated with them.

Either a context menu option to post a talent to chat or make them usable from the character sheet again to post to chat

See above

**Version**  
3.0.0

**Additional context**  
_Add any other context or screenshots about the feature request here._

**Checklist:**

dtfe:
**Type**  

- [X] Bug fix
- [X] Feature
- [ ] Refactor

**Description**  
Its technically a bug fix as we previously had this functionality so losing it in an update would be odd.
This PR aims to let users post talents and features to chat. Technically the current implementation allows anything with a description to be posted in chat, but from what I can tell nothing else uses item.use currently aside from talents, features and actions

This also covers use cases where we, for whatever reason, forgot to add an action to a talent or feature that should've been there.

**Related Issue**  
Closes #769 

**How Has This Been Tested?**  
1. Create new character
2. Add heroic path Agent to character
3. Set level to 4, insight to 3
4. Open agent talent tree
5. Give yourself Sure Outcome, Risky Behavior and Cheap Shot
6. Use all 3 in the talents section and check that they are posted to chat. Then open the ones in chat and confirm the descriptions fit
7. Open adversary starter rules
8. Go to archer
9. Post minion
10. Go to cremling swarm
11. Post swarm
12. See everything works as intended.

image

**Checklist:**  


src/system/applications/actor/components/character/talents-list.ts
```diff
@@ -233,6 +233,16 @@ const MISC_SECTION: ItemListSection = {
...
 
+    /**
+     * NOTE: Unbound methods is the standard for defining actions
+     * within ApplicationV2
+     */
+    /* eslint-disable @typescript-eslint/unbound-method */
+    static readonly ACTIONS = {
+        ...super.ACTIONS,
+        'use-item': this.onUseItem,
+    };
```
stanavdb, line 244: Missing matching `eslint-enable` rule, unnecessarily leaving `@typescript-eslint/unbound-method` disabled for the rest of the file.

src/system/documents/item.ts
stanavdb, line 1: Disagree with the chosen approach to handle this through the use() funtion. The item isn't actually being used in any meaningful sense and it muddies the responsibility of the function (and the related hook). Think the better approach would be to strip the logic of posting an item on its own out of the use function and into a separate function (something like `toChatMessage`?).

stanavdb (dismissed):

dtfe:
I agree with your thought process. It's messy to keep all that logic in item.use(), even if this is technically the item being used. 

I split it up in the item script and added some logic to the actor's useItem function to send the item as a chat message if it's not an action and has a description. If it has neither then it sends a warning in the console.

src/system/documents/item.ts
```diff
@@ -1582,6 +1531,39 @@ export class CosmereItem<
 
     /* --- Functions --- */
 
+    public async sendAsChatMessage(
```
stanavdb: Think we should rename this to something like `toChatMessage` or `asChatMessage` and then return the ChatMessage object so later code could use it if needed.

src/system/documents/item.ts
```diff
@@ -1582,6 +1531,39 @@ export class CosmereItem<
...
+                type: MESSAGE_TYPES.ACTION,
+                description: await this.getDescriptionHTML(),
+                targets: getTargetDescriptors(),
+                item: this.id,
+            },
+        };
+
+        // NOTE: Use boolean or operator (`||`) here instead of nullish coalescing (`??`),
+        // as flavor can also be an empty string, which we'd like to replace with the default flavor too
+        const flavor = undefined;
```
stanavdb: Flavor is unused and comment is irrelevant

src/system/documents/actor.ts
```diff
@@ -1104,7 +1104,16 @@ export class CosmereActor<
...
+        if (item.isAction()) {
+            return item.use({ ...options, actor: this });
+        }
+        if (item.hasDescription()) {
+            return item.sendAsChatMessage({ ...options, actor: this });
+        }
+        console.warn(
+            item.name + ' does not do anything when used. Is this intentional?',
+        );
+        return null;
```
stanavdb, line 1126: I'm not entirely against does but I'm also not entirely sure this is the right place for this kind of behaviour. Feels like it's mixing concerns (i.e. why should the Actor class control this behaviour).

stanavdb (dismissed):

src/system/documents/actor.ts
dtfe: The actor class controls the behavior as this defines what using an item on the actor sheet means. At least thats my justification for it.

stanavdb (approved):

PR #905, opened 2026-07-31
fix: modifies grabbing of item from uuid to function with compendiums

### LINKED ISSUE #904: Cant drag and drop compendium embedded actions

### Describe the bug

drag dropping a compendium embedded action does not work due to the drag and drop using fromUuidSync which isnt supported for compendium items.

### Expected Behavior

Drag drop should work as normal

### Steps To Reproduce

1. Create compendium
2. add talent in compendium
3. add embedded action to talent in compendium
4. drag action to world items
5. See error in console

### Operating System

Windows 11

### Web Browser(s)

Foundry Executable

### Hosting Method

Localhost/same PC

13.351

3.0.1

Image

dtfe:
**Type**  

- [X] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
Currently when starting to drag an embedded action in a compendium item, it runs into an error as it's using fromUuidSync and that does not work with compendium items. From what I can tell, there is no reason for _onDragStart to actually get the item itself aside from checking if said item exists. I figure just checking if the uuid exists after trying to grab it from the event is enough of a check.

After the changes, theres a new function to get embedded items which is used instead. This functions for both compendium items and regular items.

**Related Issue**  
Closes #904 

**How Has This Been Tested?**  
1. Create new item compendium
2. add talent
3. add action to talent
4. Drag embedded action in talent to world items
5. drag talent to world items
6. create folder in compendium
7. drag talent into folder
8. drag folder to world items

**Checklist:**  

### DIFF
--- a/src/system/applications/item/base.ts
+++ b/src/system/applications/item/base.ts
@@ -402,6 +402,9 @@ export class BaseItemSheet extends TabsApplicationMixin(
         const itemUuid = AppUtils.getItemUuidFromEvent(event);
         if (!itemUuid) return;
 
+        const eventItem = this.item.getEmbeddedItemFromUuid(itemUuid);
+        if (!eventItem) return;
+
         const dragData = {
             type: 'Item',
             uuid: itemUuid,
--- a/src/system/applications/item/base.ts
+++ b/src/system/applications/item/base.ts
@@ -400,12 +400,11 @@ export class BaseItemSheet extends TabsApplicationMixin(
     protected override _onDragStart(event: DragEvent) {
         // Get dragged item
         const itemUuid = AppUtils.getItemUuidFromEvent(event);
-        const item = fromUuidSync(itemUuid);
-        if (!item) return;
+        if (!itemUuid) return;
 
         const dragData = {
             type: 'Item',
-            uuid: item.uuid,
+            uuid: itemUuid,
         };
 
         // Set data transfer
--- a/src/system/documents/item.ts
+++ b/src/system/documents/item.ts
@@ -476,7 +476,12 @@ export class CosmereItem<
      */
     public getEmbeddedItemFromUuid(uuid: string): CosmereItem | null {
         const parsedUuid = foundry.utils.parseUuid(uuid);
-        if (!parsedUuid || parsedUuid.primaryId !== this.id) {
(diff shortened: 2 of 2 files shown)


stanavdb (dismissed):
The reason this check is there is also to ensure that the uuid is valid and resolves to an actual document.

dtfe:
> The reason this check is there is also to ensure that the uuid is valid and resolves to an actual document.

I understand that, but realistically if you can grab a UUID from the event it should be valid because something used item.uuid to set it there. Whether the item exists or not will be discovered when it's dropped, no? The onDragStart never actually uses the item in any meaningful way and its scope seems to be delivering an item UUID.

stanavdb:
> I understand that, but realistically if you can grab a UUID from the event it should be valid because something used item.uuid to set it there. Whether the item exists or not will be discovered when it's dropped, no? The onDragStart never actually uses the item in any meaningful way and its scope seems to be delivering an item UUID.

We can't ensure that all consumers of drop data are going to perform full validation (think 3rd party modules, for example). We shouldn't risk creating invalid drop data because we just didn't want to put in validation. Honestly even the previous validation isn't actually sufficient, as it doesn't verify the document type.

dtfe:
> > I understand that, but realistically if you can grab a UUID from the event it should be valid because something used item.uuid to set it there. Whether the item exists or not will be discovered when it's dropped, no? The onDragStart never actually uses the item in any meaningful way and its scope seems to be delivering an item UUID.
> 
> We can't ensure that all consumers of drop data are going to perform full validation (think 3rd party modules, for example). 

From what I can tell, turning this asynchronously can be dangerous as from what I can tell online, data transfer for dragevents are modifiable synchronously. Not asynchronously. So it may cause issues turning the function async. Unless theres some reason why it wouldnt in our case, I'd be happy to change it to async.

stanavdb:
> From what I can tell, turning this asynchronously can be dangerous as from what I can tell online, data transfer for dragevents are modifiable synchronously. Not asynchronously. So it may cause issues turning the function async. Unless theres some reason why it wouldnt in our case, I'd be happy to change it to async.

No I agree with you that we can't turn this async

stanavdb (approved):

PR #943, opened 2026-08-27
Fix/weapon traits behavior do not apply properly #934

### LINKED ISSUE #934: weapon traits behavior do not apply properly

### Describe the bug

ApplyDamageOptions.originatingItem does not seem to work with the embedded actions. 

Instead of getting the associated weapon with say the pierce trait, it instead gets the ephemeral document that is the strike action. Thus resulting in the system not accounting for special behaviors tied to traits for weapons.

### Expected Behavior

ApplyDamageOptions.originatingItem should return the weapon itself, not the action if it is a strike action

### Steps To Reproduce

1. Create a new character
2. set deflect to 5
3. give new character grandbow and ensure pierce is an expert trait and that the character has grandbow expertise
4. drag character token to scene
5. equip grandbow and use strike action
6. Apply damage to own character and see that deflect is applied even though pierce is on the weapon.

### Operating System

Windows 11

### Web Browser(s)

Foundry Executable

### Hosting Method

Localhost/same PC

13.351

3.0.2

_No response_

dtfe:
**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor

**Description**  
Fixes a few issues. 
One regarding how we check for pierce on the actor when applying damage. 
One issue on the item on getting itemSource where it ignores embedded items
And lastly an issue where isDefaultAction checks its own default action, instead of the parent's default action.

**Related Issue**  
Closes #934 

**How Has This Been Tested?**  
1. Create a new character
2. set deflect to 5
3. give new character grandbow and ensure pierce is an expert trait and that the character has grandbow expertise
4. drag character token to scene
5. equip grandbow and use strike action
6. Apply damage to own character and see that deflect is no longer applied on pierce applicable weapons

image

**Checklist:**  

**Additional context**  

### DIFF
--- a/src/system/documents/chat-message.ts
+++ b/src/system/documents/chat-message.ts
@@ -60,11 +60,9 @@ export class CosmereChatMessage<
     }
 
     public get itemSource(): CosmereItem | null {
-        return this.actorSource
-            ? (this.actorSource.items.get(
-                  this.getFlag(SYSTEM_ID, 'message.item'),
-              ) ?? null)
-            : null;
+        const id = this.getFlag(SYSTEM_ID, 'message.item');
+        if (!id) return null;
+        return this.actorSource?.getEmbeddedDocumentFromUuid(id) as CosmereItem;
     }
 
     public get d20Rolls(): D20Roll[] {
--- a/src/system/documents/chat-message.ts
+++ b/src/system/documents/chat-message.ts
@@ -60,9 +60,11 @@ export class CosmereChatMessage<
     }
 
     public get itemSource(): CosmereItem | null {
-        const id = this.getFlag(SYSTEM_ID, 'message.item');
-        if (!id) return null;
-        return this.actorSource?.getEmbeddedDocumentFromUuid(id) as CosmereItem;
+        const uuid = this.getFlag(SYSTEM_ID, 'message.item');
+        if (!uuid) return null;
+        return this.actorSource?.getEmbeddedDocumentFromUuid(
+            uuid,
+        ) as CosmereItem;
     }
 
     public get d20Rolls(): D20Roll[] {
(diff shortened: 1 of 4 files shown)


src/system/documents/chat-message.ts
```diff
@@ -60,11 +60,9 @@ export class CosmereChatMessage<
...
 
     public get itemSource(): CosmereItem | null {
-        return this.actorSource
-            ? (this.actorSource.items.get(
-                  this.getFlag(SYSTEM_ID, 'message.item'),
-              ) ?? null)
-            : null;
+        const id = this.getFlag(SYSTEM_ID, 'message.item');
+        if (!id) return null;
+        return this.actorSource?.getEmbeddedDocumentFromUuid(id) as CosmereItem;
```
Khunkurisu: This isn't the fault of this PR, but affects this fix so should probably be resolved here. `getEmbeddedDocumentFromUuid()` fails if a separate weapon exists, since the strike's ID is always 'ephdoc0000000000'. It does not account for partial ids being duplicates, so it grabs the first item it finds that matches.

This can be fixed by simply including the parent id head of the item id when passing it in the flags (so 'parent-id-here.ephdoc0000000000'), but my recommendation is to use `foundry.utils.randomID()` wherever in the code that ephemeral documents are generating their ID, and simply append the resulting id to 'ephdoc', which should ensure they're always unique, since incremental ids are almost always a bad idea unless they're meant to be keys (which shouldn't tie together anyway).

The latter fixes the problem by removing the possibility of overlap almost entirely, while the former simply fixes this one instance.

Khunkurisu (dismissed):

src/system/documents/chat-message.ts
Khunkurisu: You may need to generate the ID and replace the first 6 values of it with the 'ephdoc' rather than just prefixing it, as I don't know if the id has to match that length to function.

Khunkurisu (approved):

PR #957, opened 2026-09-04
fix(actor sheet): Default Resource Mode #956

### LINKED ISSUE #956: resources default mode isn't properly set

### Describe the bug

Causes issues where certain resources are stuck in the "derived" mode and can't be changed.

### Expected Behavior

Resources should default to a valid resource mode if it currently isn't set to a valid one.

### Steps To Reproduce

1. Create new adversary
2. Edit investiture
3. See that its stuck on custom but acts as though its on derived.

### Operating System

Windows 11

### Web Browser(s)

Foundry Executable

### Hosting Method

Localhost/same PC

13.351

3.1.0

_No response_

dtfe:
**Type**  

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Other (please describe): Chore

**Description**  
Fixes an issue where currently we do not properly set the default mode for resources, but do limit the available modes. Causing an issue where you can't set the resource.

It fixes the issue by making sure the current mode is one of the valid modes for the resource. If it isn't then it sets it to default which each value now has its own default mode based on whether the actor is a Character or Actor

**Related Issue**  
Closes #956 

**How Has This Been Tested?**  
1. Create new character
2. Configure all resources and see that
- Health contains Derived (default) and Override for Characters.
- Focus contains Derived (default) and Override for Characters.
- Investiture contains Override (default) for Characters
- Health contains Range (default) and Override for Adversaries.
- Focus contains Override (default) for Adversaries.
- Investiture contains Override (default) for both Adversaries.
3. See that each of these resources behave as they should according to their set mode.
4. Create new adversary
5. Repeat steps 2-3 for the adversary

Also been tested like so
1. Create new character and adversary on main branch
2. Check resource modes and see that everything is on derived by default, even if mode says something different
3. Go back to this branch, build and refresh
4. See that resource modes are properly set to the default. (See step 2 on first numbered list)

**Checklist:**  

**Additional context**  
Currently this assumes Derived.Mode.Derived is always the "default" if its a valid option. I assume this isn't changing any time soon, but if it is or we would rather prepare for that, it is easily solved by adding a "none" option in the Derived.Mode enum at the top.
"None" would then be acting as the new default for the enum which will always be invalid and thus trigger the sanitization for the resource modes and set it to the proper default.

This will be a draft PR until Khu has created a new release 3.1.1 branch which this will target.

### DIFF
--- a/src/system/data/actor/common.ts
+++ b/src/system/data/actor/common.ts
@@ -679,6 +679,80 @@ export class CommonActorDataModel<
     }
 
     /**
+     * If Resource mode is not valid for that resource, set it back to the resource's default mode.
+     * This also sets the default given Derived isn't valid and not the default.
+     */
+    private sanitizeResourceModes(): void {
+        const health = this.resources[Resource.Health].max;
+        const focus = this.resources[Resource.Focus].max;
+        const investiture = this.resources[Resource.Investiture].max;
+
+        switch (this.parent.type) {
+            case 'character':
+                // Character Health
+                if (
+                    !CONFIG.COSMERE.resources.hea.modes.character.valid.includes(
+                        health.mode,
+                    )
+                ) {
+                    health.mode =
+                        CONFIG.COSMERE.resources.hea.modes.character.default;
+                }
+                // Character Focus
+                if (
+                    !CONFIG.COSMERE.resources.foc.modes.character.valid.includes(
+                        focus.mode,
+                    )
+                ) {
+                    health.mode =
+                        CONFIG.COSMERE.resources.foc.modes.character.default;
+                }
+                // Character Investiture
+                if (
+                    !CONFIG.COSMERE.resources.inv.modes.character.valid.includes(
+                        investiture.mode,
+                    )
+                ) {
+                    investiture.mode =
+                        CONFIG.COSMERE.resources.inv.modes.character.default;
+                }
+                break;
+            case 'adversary':
+                // Adversary Health
+                if (
+                    !CONFIG.COSMERE.resources.hea.modes.adversary.valid.includes(
+                        health.mode,
+                    )
+                ) {
+                    health.mode =
+                        CONFIG.COSMERE.resources.hea.modes.adversary.default;
+                }
+                // Adversary Focus
+                if (
+                    !CONFIG.COSMERE.resources.foc.modes.adversary.valid.includes(
+                        focus.mode,
+                    )
+                ) {
+                    focus.mode =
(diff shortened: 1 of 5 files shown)


src/system/data/actor/common.ts
```diff
@@ -678,6 +678,80 @@ export class CommonActorDataModel<
         });
     }
 
+    /**
+     * If Resource mode is not valid for that resource, set it back to the resource's default mode.
+     * This also sets the default given Derived isn't valid and not the default.
+     */
+    private sanitizeResourceModes(): void {
```
Khunkurisu, line 685: This entire switch statement could be replaced with a simple object key check against the actor type, which would be far less verbose and accomplish the same thing. It loses the "oops somehow not an actor type" check, but this automatically handles new actor types and errors if we configure them poorly. Safer and easier to read.
```ts
    /**
     * If Resource mode is not valid for that resource, set it back to the resource's default mode.
     * This also sets the default given Derived isn't valid and not the default.
     */
    private sanitizeResourceModes(): void {
        const health = this.resources[Resource.Health].max;
        const healthConfig =
            CONFIG.COSMERE.resources.hea.modes[this.parent.type as ActorType];
        const focus = this.resources[Resource.Focus].max;
        const focusConfig =
            CONFIG.COSMERE.resources.foc.modes[this.parent.type as ActorType];
        const investiture = this.resources[Resource.Investiture].max;
        const investitureConfig =
            CONFIG.COSMERE.resources.inv.modes[this.parent.type as ActorType];

        health.mode = healthConfig.valid.includes(health.mode)
            ? health.mode
            : healthConfig.default;
        focus.mode = focusConfig.valid.includes(focus.mode)
            ? focus.mode
            : focusConfig.default;
        investiture.mode = investitureConfig.valid.includes(investiture.mode)
            ? investiture.mode
            : investitureConfig.default;
    }
```

Khunkurisu: The code shared here is exactly what I replaced yours with in testing and it worked perfectly fine. You're welcome to adjust it if you feel the need to for some reason, though I can't really see what that reason would be.

Khunkurisu (changes requested):
Besides the code-quality change suggested below this works fine. So once you've fixed the switch statement, this can be approved.

src/system/data/actor/common.ts
dtfe: There. You are right. This is cleaner and easier to read. I need to get better at spotting these kinds of solutions as I'm too used to just using if and switch statements and instead of ternary operators

Khunkurisu (approved):

